PR #4, opened 2016-10-07
Update for Slicer4

lassoan:
Vesselness filtering and Centerline computation modules seem to work fine. I couldn't get any useful output from Level set segmentation but maybe I used incorrect inputs.


lassoan:
The current master of vmtk/SlicerExtension-VMTK does not work at all with Slicer4, while with these changes VMTK extension builds successfully and can be installed from the extension manager on all 3 platforms (tested by temporarily changing the repository in ExtensionIndex to lassoan/SlicerExtension-VMTK).

lassoan:
Thank you!
It could make sense to transfer ownership of this project to the Slicer organization, as it would make future updates easier (it would be useful to do it for all other Slicer extensions that remain broken for a long time).

jcfr:
> It could make sense to transfer ownership of this project to the Slicer organization

Agreed.

@haehn What do you think ?

lantiga:
This is great news! Please go ahead with the transfer, let me know if you need assistance with it.

PR #5, opened 2017-03-02
The final label image results in an inverse result.

tasahi:
The original solution labeled with 5 outside the structure and 0 inside.


pieper:
Sounds reasonable to me.  Does anyone else know the code well enough to sign off and merge this?

lassoan:
Thank you, the fix has been integrated.

PR #13, opened 2018-06-06
Changed VMTK repository to vmtk/vmtk.git, rather than rlizzo/vmtk.git

mschumak:
File External_VMTK.cmake referenced the rlizzo/vmtk.git repository, which is currently 18 commits behind the actual vmtk/vmtk.git repository. Changed the git repository reference in the file.


rlizzo:
yeah thanks @mschumak! No idea why that was registered to my personal fork! Merged!

PR #17, opened 2019-07-13
COMP: Fix macOS packaging ensuring *Python.so module are correctly installed

jcfr:
Prior to this commit the vtk*Python.so modules where installed in

  lib/Slicer-4.11/qt-loadable-modules/Python

Following this commit, modules are installed in the correct directory
allowing them to be fixed up:

  lib/Slicer-4.11/qt-loadable-modules/

See https://discourse.slicer.org/t/vmtk-vessel-filtering-not-working/1263/26?u=jcfr


lassoan:
Awesome, thanks a lot!

pieper:
I confirmed the new package works, but there was still a small python error.  With [this fix](https://github.com/vmtk/SlicerExtension-VMTK/commit/df945a4ffe75e1e9f9de83b330a2914853915948) vesselness filtering works for me on mac.

PR #18, opened 2020-07-08
Proposed module CenterLineMetrics for this collection of tools.

chir-set:
It allows to plot the average diameter distribution along a centerline.
The graph may show the cumulative length from an endpoint, or the projected
location on a selected axis.

Please see :
https://discourse.slicer.org/t/vmtk-centreline-and-metrics/12203


pieper:
Sounds really nice - I would be able to make use of this 👍

CenterLineMetrics/CenterLineMetrics.py
```diff
@@ -0,0 +1,285 @@
...
+  def __init__(self, parent):
+    ScriptedLoadableModule.__init__(self, parent)
+    self.parent.title = "Centerline metrics"  # TODO: make this more human readable by adding spaces
+    self.parent.categories = ["Vascular Modeling Toolkit"]  # TODO: set categories (folders where the module shows up in the module selector)
+    self.parent.dependencies = []  # TODO: add here list of module names that this module requires
+    self.parent.contributors = ["SET (Hobbyist)"]  # TODO: replace with "Firstname Lastname (Organization)"
+    self.parent.helpText = """
+This module plots average diameters around a VMTK centerline model. It is intended for non-bifurcated centerlines.
+"""  # TODO: update with short description of the module
+    self.parent.helpText += self.getDefaultModuleDocumentationLink()  # TODO: verify that the default URL is correct or change it to the actual documentation
```
lassoan, line 30: Replace this with actual link to the module documentation

CenterLineMetrics/CenterLineMetrics.py
```diff
@@ -0,0 +1,285 @@
...
+  requiring an instance of the Widget.
+  Uses ScriptedLoadableModuleLogic base class, available at:
+  https://github.com/Slicer/Slicer/blob/master/Base/Python/slicer/ScriptedLoadableModule.py
+  """
+  def __init__(self):
+    self.inputModelNode = None
+    self.outputPlotSeriesNode = None
+    self.outputTableNode = None
+    self.plotChartNode = None
+    self.axis = 2 # Default to vertical
```
lassoan, line 161: this is not needed anymore

CenterLineMetrics/CenterLineMetrics.py
```diff
@@ -0,0 +1,285 @@
...
+    if not self.ui.outputPlotSeriesSelector.currentNode():
+      outputPlotSeriesNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLPlotSeriesNode")
+      self.ui.outputPlotSeriesSelector.setCurrentNode(outputPlotSeriesNode)
+
+  def onApplyButton(self):
+    self.createOutputNodes()
+    self.logic.run()
+  
+  def onRadioR(self):
+    self.logic.setAxis(0)
```
lassoan, line 125: remove the axis selectors, they are not needed anymore

chir-set:
The axis selector is important from a clinical point of view.

We can move the 2D view slider bar to show that plane where the diameter is minimal for a stenosis, or where it is maximal for an aneurysm. The projected position on the axis points to these 2D view locations.  We lose this information with the cumulative distance. Else, for a physician, the plot would be of limited interest. (Ideally, the 2D view location should update at right angle to the centerline as we CTRL+MouseMove along the plot's X axis. But that's another process, and well above my skills).

Now, we can choose to default to distance along curve.

CenterLineMetrics/CenterLineMetrics.py
chir-set: Will search how to do it all.

chir-set: Done.

lassoan:
Ok, this information is very useful. We need to talk about the clinical need. 

Both stenosis and aneurysm cross-sections can be non-circular, so reporting a single diameter would be limited, potentially misleading information. So, I think we should take advantage of curved planar reformat module and the feature that I added recently, which allows mapping between positions in the original and straightened vessels.

Can you tell a bit more about your workflow? What measurements and visualizations you would like to have?

chir-set:
Well, in surgery, we need to evaluate stenosis degree for occlusive disease and how diffuse they are along an artery.  This information is visually evident in a segmentation model, or even in volume rendering very often, and a decision can be readily taken. We won't derive centerlines everyday and study the diameter distribution. But it will certainly be an additional helpful tool in selected cases. Beyond pure clinical need, these models, lines and plots, can simplify information delivery in group talks, or as support material for students.

For aneurysms, the maximum diameter including thrombus is of paramount importance. Here, the diameter reported by the centerline approach will rarely be a decision cornerstone. Yet, the diameters (and morphology) above and below the renal arteries are of capital importance during surgery planning.

A cross-section perpendicular to the centerline while moving along the centerline is helpful. The curved planar reformat module is much adapted to always show this type of cross-section. I have played with it and it is helpful. It's main inconvenience is that the reformatted volume implicitly loses interpretable anatomical references around the straight line. It's not a big problem as we know we are dealing with a cut-off volume; we should not expect more than what we are trying to do (studying the artery).

I am playing with the Reformat module right now. The intent is to move the reformat widget's center along the centerline and use SetSliceNormal() to get the slice perpendicular to the centerline. Nothing concrete yet, will push an experimental module if anything worthwhile comes out of it.

As for this particular module, I can default to cumulative distance mode if you wish, and hide the relevant options in a collapsible pane.

I may also donate this module to Slicer's team, if this has any meaning.

Regards.

chir-set:
UI and defaults are updated as such : 

Hide distance mode and axis selection in a collapsible button.
Default to cumulative distance mode.

lassoan:
Thanks for the detaile description.

>  It's main inconvenience is that the reformatted volume implicitly loses interpretable anatomical references around the straight line

Recent versions of the module place the reformatted volume in the same position and orientation as the original (as much as possible), so anatomical referencing is much easier. I've also completely reworked the reformatting recently so that there is an invertible transform between the original and straightened volume: you can transform positions (models, markups, etc.) to/from the straightened image. See this video for an example: https://www.youtube.com/watch?v=WJQexDTiRRc. For example, if you mark a point in the straightened segment then you can automatically get the position in the original image. Currently, you need to switch between transform and markups modules to take advantage of this feature, but it would not be hard to add two-way synchronization between two markup fiducial lists. You can also interact with plots, so you could jump the slice to the position where you clicked in the diameter plot.

Anyway, we can get this integrated as is and add more features later.

chir-set:
I've learned new things on your video and have to study it more.

Thanks.

chir-set:
Since the merge in SlicerExtention-VMTK, the .ui file is probably altered.

Qt Designer does not show any signal-slot connection. And indeed, when the module is selected in Slicer, all three MRML node comboboxes are disabled.

Can you please have a look ?

Thanks.

chir-set:
> Since the merge in SlicerExtention-VMTK, the .ui file is probably altered.
> 

This is due to 4c2ed041 :

In the .ui file,

`<widget class="qMRMLWidget" name="CenterlineMetrics">`

has replaced 

`<widget class="qMRMLWidget" name="CenterLineMetrics">`

But, 'CenterLineMetrics' still persists in many places, like

```
<class>CenterLineMetrics</class>
<sender>CenterLineMetrics</sender>
```

chir-set:
> This is due to [4c2ed04](https://github.com/vmtk/SlicerExtension-VMTK/commit/4c2ed041ec4847e5f12907ce3302d4bed5cf3ed2) :
> 

A simple fix is 

`sed -i "s/CenterLineMetrics/CenterlineMetrics/g" CenterlineMetrics.ui`

It works normally afterwards.

lassoan:
I've replaced (nearly) all instances of CenterLine by Centerline for consistency, but apparently missed some. Could you please send a pull request with the fix?

PR #19, opened 2020-07-20
Complete rename CenterLineMetrics to CenterlineMetrics.

chir-set:
Four lines with 'CenterLineMetrics' persisted in CenterlineMetrics.ui.
This would disable all 3 MRML node comboboxes.


lassoan:
Thank you!

PR #24, opened 2021-01-20
COMP: Use ExtractCenterline module in custom apps

cpinter:
In the custom app superbuild the variables Slicer_VERSION_MAJOR and Slicer_VERSION_MINOR are both 0 (although the Slicer_VERSION_FULL variable is correct). To fix this problem, we assume that if the version is 0 then 4.13 or later is used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,7 @@ endif()
 #-----------------------------------------------------------------------------
 add_subdirectory(VesselnessFiltering)
 add_subdirectory(LevelSetSegmentation)
-if(${Slicer_VERSION_MAJOR}.${Slicer_VERSION_MINOR} VERSION_GREATER_EQUAL 4.11)
+if(${Slicer_VERSION_MAJOR}.${Slicer_VERSION_MINOR} VERSION_GREATER_EQUAL 4.11 OR ${Slicer_VERSION_MAJOR} EQUAL 0)
   # New-generation centerline computation - only available for recent Slicer versions
   add_subdirectory(ExtractCenterline)
 else()


CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ endif()
 #-----------------------------------------------------------------------------
 add_subdirectory(VesselnessFiltering)
 add_subdirectory(LevelSetSegmentation)
-if(${Slicer_VERSION_MAJOR}.${Slicer_VERSION_MINOR} VERSION_GREATER_EQUAL 4.11)
+if(${Slicer_VERSION_MAJOR}.${Slicer_VERSION_MINOR} VERSION_GREATER_EQUAL 4.11 OR ${Slicer_VERSION_MAJOR} EQUAL 0)
```
jcfr, line 40: I just tried to add the statement `message("Slicer_VERSION_MINOR:${Slicer_VERSION_MINOR}")` into the CMakeLists.txt of a custom app module I generated yesterday, and it displays the appropriate value.

Out of curiosity, how is the module integrated ?

cpinter: In the main CMakeLists.txt of the custom app we have

```
# SlicerVMTK
set(extension_name "SlicerVMTK")
set(${extension_name}_SOURCE_DIR "${CMAKE_BINARY_DIR}/${extension_name}")
FetchContent_Populate(${extension_name}
 SOURCE_DIR     ${${extension_name}_SOURCE_DIR}
 GIT_REPOSITORY ${EP_GIT_PROTOCOL}://github.com/vmtk/SlicerExtension-VMTK.git
 GIT_TAG        3787ea4a300da28ec5f0824f0715f2713b631155
 GIT_PROGRESS   1
 QUIET
 )
list(APPEND Slicer_EXTENSION_SOURCE_DIRS ${${extension_name}_SOURCE_DIR})
```

lassoan:
There are two sets of variables - for the Slicer platform and the main project (custom application):

  Slicer_BUILDDATE
  Slicer_VERSION
  Slicer_VERSION_FULL
  Slicer_REVISION
  Slicer_WC_REVISION
  Slicer_WC_REVISION_HASH
  Slicer_WC_URL

  Slicer_MAIN_PROJECT_BUILDDATE
  Slicer_MAIN_PROJECT_VERSION
  Slicer_MAIN_PROJECT_VERSION_FULL
  Slicer_MAIN_PROJECT_REVISION
  Slicer_MAIN_PROJECT_WC_REVISION
  Slicer_MAIN_PROJECT_WC_REVISION_HASH
  Slicer_MAIN_PROJECT_WC_URL

When you develop a custom application, the version has to be specified in [slicer-application-properties.cmake](https://github.com/KitwareMedical/SlicerCustomAppTemplate/blob/master/%7B%7Bcookiecutter.project_name%7D%7D/Applications/%7B%7Bcookiecutter.app_name%7D%7DApp/slicer-application-properties.cmake#L6-L14) and should not be necessary to override the platform version.

@cpinter please check if this fixes the problem for you. If it does then maybe we can remove the platform version override mechanism to make things simpler.

jcfr:
Here is the issue:

Integrate of extension source directories is done [here](https://github.com/Slicer/Slicer/blob/1c6eb5480729daa802e0028898d27a5ff2a9e114/CMakeLists.txt#L1013-L1084)

To support this, a custom `SlicerConfig.cmake` has to be generated to make sure the `find_package(Slicer)` call in the extension is a no-op.

It turns out that the generated `SlicerConfig.cmake` set these variables to `0`

To fix the problem, we simply need to make sure the `SlicerConfig.cmake` file configured [here](https://github.com/Slicer/Slicer/blob/1c6eb5480729daa802e0028898d27a5ff2a9e114/CMakeLists.txt#L1016-L1019) has the expected values for `Slicer_VERSION_*` variables

jcfr:
PR will be submitted shortly

jcfr:
See https://github.com/Slicer/Slicer/pull/5396

cc: @cpinter

jcfr (changes requested):
Fix has been integrated upstream, instead of applying this change, I suggest to close this PR and update the version of Slicer in your custom application.

cpinter:
Sorry had a meeting. Thanks @jcfr it makes good sense. I'll update the underlying Slicer in the custom app and see if the variables have non-zero values.

@lassoan
> "There are two sets of variables"

Yes I know this but the two sets are used for different things, and SlicerVMTK made a decision about the centerline module to use based on the Slicer version which was unavailable.

cpinter:
@jcfr @lassoan I can confirm that I have the correct variable contents using the fix. I'm closing this PR. Thanks a lot for the quick help!

PR #27, opened 2021-05-24
Add RAS point coordinates to centerline model table.

chir-set:
In addition to distance and diameter columns, three more columns R, A, S
are added, corresponding to the coordinates of each point of the
centerline model.

Following https://discourse.slicer.org/t/veselness-filter/17739

The user requesting this has not specified its usefullness. It seems harmless though.

### DIFF
--- a/CenterlineMetrics/CenterlineMetrics.py
+++ b/CenterlineMetrics/CenterlineMetrics.py
@@ -205,6 +205,9 @@ class CenterlineMetricsLogic(ScriptedLoadableModuleLogic):
     # Create arrays of data
     distanceArray = self.getArrayFromTable(outputTable, DISTANCE_ARRAY_NAME)
     diameterArray = self.getArrayFromTable(outputTable, DIAMETER_ARRAY_NAME)
+    rArray = self.getArrayFromTable(outputTable, R_ARRAY_NAME)
+    aArray = self.getArrayFromTable(outputTable, A_ARRAY_NAME)
+    sArray = self.getArrayFromTable(outputTable, S_ARRAY_NAME)
 
     # From VMTK README.md
     points = slicer.util.arrayFromModelPoints(inputModel)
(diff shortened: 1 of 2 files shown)


lassoan (commented):
This looks great, thank you!

Just two things would need to be improved:
- We use LPS coordinates in files. Could you add an option to choose between RAS and LPS coordinates?
- In most places we store coordinates in 3 components of a single vtkDoubleArray. You can keep the 3 separate arrays as an option.

chir-set:
> * store coordinates in 3 components of a single vtkDoubleArray

You mean, one column with cell values as 

[-10.717971  18.031818 -83.89904 ]
[-10.72737   17.982111 -84.08649 ]
[-10.748347  17.931984 -84.34901 ]

?

lassoan:
Yes, one vtkDoubleArray with number of components set to 3. For example, segment statistics module uses it for oriented bounding box position and orientation.

chir-set:
> * We use LPS coordinates in files.
> * we store coordinates in 3 components of a single vtkDoubleArray. 

@lassoan 
Point coordinates are now presented in the table as LPS or RAS values, in 1 single or 3 separate columns.

It would be nice to fit the column width when presented in a single column, but I don't have a clue how to do that.

lassoan:
Thank you! I've squashed and refactored the code a bit and merged it in https://github.com/vmtk/SlicerExtension-VMTK/commit/72a52b24b14ec70b03061e89075ab3cd018c30a6.

chir-set:
> I've squashed and refactored the code a bit and merged it in [72a52b2](https://github.com/vmtk/SlicerExtension-VMTK/commit/72a52b24b14ec70b03061e89075ab3cd018c30a6).

@lassoan 
Sorry to bother yo again, but the 'Projected' mode is completely ignored following the refractoring. The table and plot are not updated in this mode, whichever axis is selected.

PR #28, opened 2021-06-07
Fix distance mode locked on default cumulative.

chir-set:
Remove non-existing widgets.
@lassoan


lassoan (approved):
Thanks a lot for testing and fixing this!

PR #29, opened 2021-06-12
Deprecate projected mode. Add helper functions to move slices.

chir-set:
Please consider the described enhancements for merging.

Concern Centerline metrics and Cross-section analysis modules.

Regards.

### DIFF
--- a/CenterlineMetrics/CenterlineMetrics.py
+++ b/CenterlineMetrics/CenterlineMetrics.py
@@ -75,11 +75,6 @@ class CenterlineMetricsWidget(ScriptedLoadableModuleWidget, VTKObservationMixin)
     self.ui.inputModelSelector.connect("currentNodeChanged(vtkMRMLNode*)", self.onSelectNode)
     self.ui.outputPlotSeriesSelector.connect("currentNodeChanged(vtkMRMLNode*)", self.onSelectNode)
     self.ui.outputTableSelector.connect("currentNodeChanged(vtkMRMLNode*)", self.onSelectNode)
-    self.ui.radioR.connect("clicked()", self.onRadioR)
-    self.ui.radioA.connect("clicked()", self.onRadioA)
-    self.ui.radioS.connect("clicked()", self.onRadioS)
-    self.ui.radioCumulative.connect("clicked()", self.onRadioCumulative)
-    self.ui.radioProjected.connect("clicked()", self.onRadioProjected)
     self.ui.radioLPS.connect("clicked()", self.onRadioLPS)
     self.ui.radioRAS.connect("clicked()", self.onRadioRAS)
     self.ui.distinctColumnsCheckBox.connect("clicked()", self.onDistinctCoordinatesCheckBox)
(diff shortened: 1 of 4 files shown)


CrossSectionAnalysis/CrossSectionAnalysis.py
```diff
@@ -588,6 +613,20 @@ def selectCurveInMarkupsModule(self):
     else:
         markupsWidget.setEditedNode(None)
 
+# Convenience function to get the point of minimum or maximum diameter. Is useful for arterial stenosis (minimum) or aneurysm (maximum).
+  def getExtremeDiameterPoint(self, boolMaximum):
+    if self.inputSliceNode is None or self.inputPath is None or (self.vmtkCenterlineRadii.size == 0):
+        return -1
+    target = -1
+    for i in range(self.vmtkCenterlineRadii.size):
```
lassoan, line 621: You can get index of max/min value using numpy - see https://thispointer.com/find-max-value-its-index-in-numpy-array-numpy-amax/. It would much faster and less code.

CenterlineMetrics/CenterlineMetrics.py
```diff
@@ -75,11 +75,6 @@ def setup(self):
     self.ui.inputModelSelector.connect("currentNodeChanged(vtkMRMLNode*)", self.onSelectNode)
     self.ui.outputPlotSeriesSelector.connect("currentNodeChanged(vtkMRMLNode*)", self.onSelectNode)
     self.ui.outputTableSelector.connect("currentNodeChanged(vtkMRMLNode*)", self.onSelectNode)
-    self.ui.radioR.connect("clicked()", self.onRadioR)
```
lassoan, line 78: I always felt that this was just a workaround and should not be necessary. It is great that this will be gone.

CenterlineMetrics/CenterlineMetrics.py
```diff
@@ -127,7 +138,51 @@ def onRadioRAS(self):
...
+  def moveSliceView(self, value):
+    self.logic.moveSliceView(self.ui.sliceViewSelector.currentNode(), value)
+    self.updateUIWithMetrics(value)
+  
+  def updateUIWithMetrics(self, value):
+    tableNode = self.logic.outputTableNode
+    if tableNode is None:
+        return
+    coordinateArray = self.logic.getRASCoordinatesAtIndex(value)
+    distance = str(round(tableNode.GetTable().GetValue(int(value), 0).ToDouble(), 2)) + " mm"
```
lassoan, line 154: It would be better not to hardcode the unit, but retrieve the formatted length string from the unit node:

```python
>>> selectionNode = slicer.mrmlScene.GetNodeByID("vtkMRMLSelectionNodeSingleton")
>>> unitNode = slicer.mrmlScene.GetNodeByID(selectionNode.GetUnitNodeID("length"))
>>> unitNode.GetDisplayStringFromValue(34.1)
' 34.1000 mm'
```

lassoan (commented):
These changes are great, thank you. Just added a few minor comments.

CenterlineMetrics/CenterlineMetrics.py
chir-set: I could not get to a similar trick to format orientation in degrees. Is there a key to get an orientation unit node ?

chir-set:
@lassoan 

Could you please review the corrected code ? (It's clear you're quite busy.)

Thanks.

CenterlineMetrics/CenterlineMetrics.py
lassoan: Since angle is always between 0-360 and unit is always degree, there is no unit node for angle by default.

lassoan:
Thank you, merged!

PR #32, opened 2021-07-08
Centerline metrics : show cross-section model and measure its area

chir-set:
Please consider this patch for merging. @lassoan

Regards.
############################################################################

This patch shows an exact-fit cross-section of a segment along a centerline in a 3D view, and makes its surface area available.

It works on a closed representation of a segment, which is plane cut at an arbitrary point of the centerline. The segment points of the cut plane are then ordered by proximity between points. These proximities allow to reject segment parts that are detached from the island around the centerline. The remaining points are used to create a model, which is thus the cross-section of the segment. Getting the surface area of this model is then straight-forward.

In addition, the diameter of a virtual circle having this area is also shown, as well as the difference to the VMTK computed diameter. (This may help to better understand some arterial stenosis.)

As a UI enhancement, a push button allows toggling between the previous and the plot layout.

N.B : a volume node is not needed.

### DIFF
--- a/CenterlineMetrics/CenterlineMetrics.py
+++ b/CenterlineMetrics/CenterlineMetrics.py
@@ -72,6 +72,10 @@ class CenterlineMetricsWidget(ScriptedLoadableModuleWidget, VTKObservationMixin)
     self.resetMoveToPointSliderWidget()
     self.ui.advancedCollapsibleButton.checked = False
     self.setUnitNodePrecision()
+    self.ui.toggleLayoutButton.visible = False
+    self.previousLayoutId = slicer.app.layoutManager().layout
+    self.ui.segmentationCollapsibleButton.collapsed = True
+    self.ui.coordinatesCollapsibleButton.collapsed = True
 
     # connections
     self.ui.applyButton.connect('clicked(bool)', self.onApplyButton)
@@ -85,6 +89,10 @@ class CenterlineMetricsWidget(ScriptedLoadableModuleWidget, VTKObservationMixin)
     self.ui.jumpCentredInSliceNodeCheckBox.connect("clicked()", self.onJumpCentredInSliceNodeCheckBox)
     self.ui.moveToMinimumPushButton.connect("clicked()", self.moveSliceViewToMinimumDiameter)
     self.ui.moveToMaximumPushButton.connect("clicked()", self.moveSliceViewToMaximumDiameter)
+    self.ui.toggleLayoutButton.connect("clicked()", self.toggleLayout)
+    self.ui.segmentationSelector.connect("currentNodeChanged(vtkMRMLNode*)", self.onSelectSegmentationNodes)
+    self.ui.segmentSelector.connect("currentSegmentChanged(QString)", self.onSelectSegmentationNodes)
+    self.ui.showAvailableCrossSectionsButton.connect("clicked()", self.onShowAvailableCrossSections)
 
     # Refresh Apply button state
     self.onSelectNode()
@@ -94,23 +102,10 @@ class CenterlineMetricsWidget(ScriptedLoadableModuleWidget, VTKObservationMixin)
     Called when the application closes and the module widget is destroyed.
     """
     self.removeObservers()
-    
-  def onApplyButton(self):
-    """
-    Run processing when user clicks "Apply" button.
-    """
-    try:
-      self.createOutputNodes()
-      self.logic.run()
-
-    except Exception as e:
-      slicer.util.errorDisplay("Failed to compute results: "+str(e))
-      import traceback
-      traceback.print_exc()
   
   def onSelectNode(self):
-    self.ui.applyButton.enabled = self.ui.inputModelSelector.currentNode()
     self.logic.setInputModelNode(self.ui.inputModelSelector.currentNode())
+    self.ui.applyButton.enabled = self.logic.isInputModelValid()
     self.logic.setOutputTableNode(self.ui.outputTableSelector.currentNode())
     self.logic.setOutputPlotSeriesNode(self.ui.outputPlotSeriesSelector.currentNode())
     self.ui.moveToPointSliderWidget.setValue(0)
@@ -126,10 +121,17 @@ class CenterlineMetricsWidget(ScriptedLoadableModuleWidget, VTKObservationMixin)
       self.ui.outputPlotSeriesSelector.setCurrentNode(outputPlotSeriesNode)
 
   def onApplyButton(self):
+    if not self.logic.isInputModelValid():
+        msg = "Input is invalid."
+        slicer.util.showStatusMessage(msg, 3000)
+        logging.info(msg)
+        return # Just don't do anything
     self.resetUIWithEmptyMetrics()
+    self.resetSurfaceAreaUIWithEmptyMetrics()
     self.createOutputNodes()
     self.logic.run()
-    self.ui.moveToPointSliderWidget.maximum = self.logic.outputTableNode.GetTable().GetNumberOfRows() - 1
+    self.ui.moveToPointSliderWidget.maximum = self.logic.outputTableNode.GetTable().GetNumberOfRows() -1 -1 
+    self.ui.toggleLayoutButton.visible = True
     
   def onRadioLPS(self):
     self.logic.coordinateSystemColumnRAS = False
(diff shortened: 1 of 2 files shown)


CenterlineMetrics/CenterlineMetrics.py
```diff
@@ -382,11 +443,91 @@ def getExtremeDiameterPoint(self, boolMaximum):
...
+    util.islandModelNode = self.islandModelNode
+    
+    # Place a plane perpendicular to the centerline
+    plane = vtk.vtkPlane()
+    plane.SetOrigin(center)
+    plane.SetNormal(normal)
+    
+    # Work on the segment's closed surface
+    closedSurfacePolyData = vtk.vtkPolyData()
+    self.segmentationNode.GetClosedSurfaceRepresentation(self.currentSegmentID, closedSurfacePolyData)
```
lassoan, line 467: The closed surface representation must be transformed to the world coordinate system. You can get the transform using `GetTransformBetweenNodes` see complete example in CombineModels.py.

CenterlineMetrics/CenterlineMetrics.py
```diff
@@ -382,11 +443,91 @@ def getExtremeDiameterPoint(self, boolMaximum):
...
+    if planePoints.GetNumberOfPoints() < 3:
+        logging.info("Not enough points to create surface")
+        return
+    
+    # Sort the points by proximity to retain those around the centre of the centerline
+    vCenter = [center[0], center[1], center[2]]
+    planePointsCopy = vtk.vtkPoints()
+    planePointsCopy.DeepCopy(planePoints)
+    # Empties planePointsCopy
+    sortedPoints = util.sortPointsByProximity(planePointsCopy, vCenter)
```
lassoan, line 493: Could you use vtkConnectivityFilter's ExtractionModeToClosestPointRegion method instead?

https://vtk.org/doc/nightly/html/classvtkConnectivityFilter.html#a0afa68eb6a4aded83fbc8c0951ac5a69

CenterlineMetrics/CenterlineMetrics.py
```diff
@@ -382,11 +443,91 @@ def getExtremeDiameterPoint(self, boolMaximum):
...
+    # Make a model out of the points retained
+    islandPolyData = vtk.vtkPolyData()
+    # Prefer an inverted color for the model
+    segmentColor = currentSegment.GetColor()
+    crossSectionColor = [1 - segmentColor[0], 1 - segmentColor[1], 1 - segmentColor[2]]
+    # Finally create the model
+    self.currentSurfaceArea = self.xFitSurfaceProjectWarp(islandPoints, islandPolyData, crossSectionColor, currentSegment.GetName())
+
+  # Create an exact-fit model representing the cross-section.
+  def xFitSurfaceProjectWarp(self, islandPoints, islandPolyData, color = [1, 1, 0], segmentName = ""):
```
lassoan, line 505: We only need these complicated warpings if the curve is non-planar. If the curve is planar then https://vtk.org/doc/nightly/html/classvtkContourTriangulator.html is a much better choice. The main advantage of this filter is that it can handle not just concave regions but also if there are holes in the cross-section (e.g., you may have a small bubble in the segmentation). For proper handling of holes, you need to first triangulate the cross-section and then extract the part that is closest to the the centerline.

CenterlineMetrics/CenterlineMetrics.py
```diff
@@ -144,27 +145,55 @@ def onJumpCentredInSliceNodeCheckBox(self):
...
+    distance = self.logic.getUnitNodeDisplayString(tableNode.GetTable().GetValue(int(value), 0).ToDouble(), "length")
+    diameter = self.logic.getUnitNodeDisplayString(tableNode.GetTable().GetValue(int(value), 1).ToDouble(), "length")
     coordinate = "R " + str(round(coordinateArray[0], 1))
     coordinate += ", A " + str(round(coordinateArray[1], 1))
     coordinate += ", S" + str(round(coordinateArray[2], 1))
     self.ui.distanceValueLabel.setText(distance)
     self.ui.diameterValueLabel.setText(diameter)
     self.ui.coordinatesValueLabel.setText(coordinate)
+  
+  # Show the surface area. In addition, derive the diameter of the circle having this area. Show also the difference and the ratio to the VMTK diameter.
```
lassoan, line 169: Extraction of the cross-section might not be always trivial. To allow the user to check what exact shape was used for computing the cross-sections, it would be nice to have an advanced option to generate a model that contains all the cross-sections.

lassoan (commented):
Thank you, this will be useful. I've added a couple of comments inline.

CenterlineMetrics/CenterlineMetrics.py
chir-set: > The closed surface representation must be transformed to the world coordinate system. You can get the transform using `GetTransformBetweenNodes`

Ok, one thing at a time.
While I can broadly see the problem in the other changes to make, here I'm clueless.

In fact, the "[World coordinate system](https://www.slicer.org/wiki/Coordinate_systems#World_coordinate_system)" is quite obscure to me. I suppose it's the normal XY axis with (0,0) origin being the lower left corner.
Is it self.segmentationNode that should be transformed to world coordinates ? Really, I'm clueless how to use GetTransformBetweenNodes to do that.

And why should we do that?

Thanks for your replies.

lassoan: "World" coordinate system, also known as "Renderer" or "RAS" coordinate system is the space where nodes are, if no transforms are applied to them. If you apply a transform to a node (e.g., Curved Planar Reformat module computes a straightening transform) then it does not change the point coordinates in a transformed model, just uses the transform to translate/rotate/warp the model for display. Some more information is provided in [Transforms module documentation](https://slicer.readthedocs.io/en/latest/user_guide/modules/transforms.html), but if anything is not clear then let me know and I'll either clarify here or add some more details to the documentation.

The module must not return incorrect results if a parent transform is set to the centerline and/or the segmentation. Ideally, you would apply the transform to the nodes to the world coordinate system, but you may postpone this for a while and for now just report an error if a parent transform set and instruct the user to harden or remove the parent transforms.

chir-set: > The module must not return incorrect results if a parent transform is set to the centerline and/or the segmentation

A segmentation transform is now applied to the model. I suppose that if the user applies different transforms to the segmentation and the centerline, he knows what he is doing.

CenterlineMetrics/CenterlineMetrics.py
chir-set: > vtkConnectivityFilter's ExtractionModeToClosestPointRegion method

It is now used to select the island around the centerline. But the points still need to be rearranged by proximity, else the code becomes a CPU burner !

lassoan: The slice intersection should not contain more than a thousand points (more likely a few hundred). Any C++ code would do any processing on such small number of points in a fraction of a millisecond. Do you mean that there is still some processing that you would need to do in Python?

chir-set: > should not contain more than a thousand points

The maximum number of points per section is 168 in my test data.

If I just pass the output of connectivityFilter.GetPolyDataOutput().GetPoints() to FitSurfaceProjectWarp(), a deadly loop is triggered, eating all RAM and CPU. If the points are sorted by proximity, then the model is fastly created.

Now you advise to use vtkContourTriangulator instead of FitSurfaceProjectWarp(). That's perhaps the issue. I'll try that and get back.

CenterlineMetrics/CenterlineMetrics.py
lassoan: We cannot assume that the user knows what he is doing. User expectation is that things are foolproof.

You can also easily transform the centerline model (https://github.com/PerkLab/SlicerSandbox/blob/db2fea8b59bfc30209740c546892c998f3adb169/CombineModels/CombineModels.py#L324-L329) and either use VTK API to get the coordinates from the transformed polydata or get the coordinate array as a numpy array (https://github.com/Slicer/Slicer/blob/5f6800e7de2ddc917ad820f63daf60a4a6ea65e2/Base/Python/slicer/util.py#L1474-L1476).

CenterlineMetrics/CenterlineMetrics.py
lassoan: FitSurfaceProjectWarp expects the points in order. If you are lucky then sorting by distance is similar enough, but in general you cannot rely on this. Probably if you create models from the cross-sections then you'll see errors if you order polygon points based on distance. Anyway, vtkContourTriangulator takes care of this, too, as it uses connectivity between points to order them correctly.

chir-set: I understand the advantages of using vtkContourTriangulator.
I am unable to get a surface model with : 

contourTriangulator = vtk.vtkContourTriangulator()
islandPolyData.SetPoints(islandPoints)
islandPolyData.Modified()
contourTriangulator.SetInputData(islandPolyData)
contourTriangulator.Update()
slicer.modules.models.logic().AddModel(contourTriangulator.GetOutput())

Calling contourTriangulator.Triangulate is worse, leading to a crash.

lassoan: Contour triangulator requires the entire cross-section of the polygon, not just the points.

chir-set: > entire output of the cutter (including cells - connectivity between points), not just the points
This did the trick, now it works, thanks.

CenterlineMetrics/CenterlineMetrics.py
chir-set: > User expectation is that things are foolproof.

That's a big problem : the user moves the segmentation to the left, and moves the centerline to the left plus a rotation - what are we supposed to do ? IMO, users are responsible for what they do.

lassoan: I just mean that we need to need to do all the measurements on what we display. Ability to display the triangulated cross-sections will make it absolutely clear what surfaces are measured.

CenterlineMetrics/CenterlineMetrics.py
chir-set: > generate a model that contains all the cross-sections

MarkupsToModel, fed with the contour points of all cross-sections, will not fit the bill, as it used vtkDelaunay3D and this is bad for concave surfaces.

I've searched a little in VTK docs and could not find a triangulator/filter that would build a 3D surface at all. May be you could hint at one tool, or elaborate your view.

Because this would amount to recreating a part of the segment, and I'm wondering about the exact usefulness.

We can always create as many static cross-section models as there are points in the centerline, that would make a confusing UI. I don't think that's what you mean though.

Using 'Combine models' ? Would it interpolate nicely between successive sections ?

Waiting for your input.

lassoan: Your can simply append each cross-section (that you use for area computation) into a single vtkPolyData using vtkAppendPolyData filter.

chir-set: Using : 

self.appendedSections.AddInputData(islandPolyData)
self.appendedSections.Update()

on the first few points, the result is as a discontinued pile of chips : 

Screenshot

Not very appealing on first view. Is it OK for you ? Or show we offer the possibility to stack sections at points chosen by the user only, and not all sections ? Once in vtkAppendPolyData, it won't be removable.

lassoan: This is good, it already allows you to spot any errors in the cross-section extraction. If you want to allow the user to see only one or a selected range then you can assign scalars to points (the same value for every point of a slice, the value can be the slice index). Then you can use Thresholding feature in Models module's Scalars section to choose which ones to display.

chir-set: OK, I see what you mean. I'll venture into that ASAP (other obligations appeared, then summer holidays).

lassoan: If you commit your latest code then I can finalize and merge it, just let me know when you are done.

chir-set: Hopefully, I should be able to finalize it in 2-3 days.

chir-set: I should be able to push a commit within 24 H.

lassoan: No problem at all, 2-3 days is completely fine, take your time. (I though I would jump in if you were to put this on hold for several weeks).

chir-set: @lassoan 
Please review my last commit.
I don't plan to add more functions for now, but have a few days to fix what you deem necessary.
Thanks.

chir-set: Hi @lassoan ,

Could you review the last commit ? If you are on summer vacation, there's nothing compelling right away.

Regards.

lassoan:
There are a number of things to improve, but they should not prevent merging this. I've rebased and squashed all the commits and will merge it now. I'll submit a ticket to track the fixes/enhancements.

Issue #36, opened 2021-08-09
Merge Centerline metrics and Cross-section analysis modules


lassoan:
There is significant overlap between features of Centerline metrics and Cross-section analysis modules. The two modules should be merged into one module to combine their strengths and reduce maintenance workload.

lassoan:
@chir-set it would be great if you could work on this, too.

chir-set:
I agree, will be done ASAP, won't be in the coming days however.

chir-set:
Hi @lassoan ,

I'm working on 'Centerline metrics' again to make it work with a freely drawn markups curve, one that does not originate from 'Extract centerline'. I hope it's the final touch in the process of merging 'Cross-section analysis'.

Please review the code [here](https://github.com/chir-set/SlicerExtension-VMTK).

Basically, it adds a 'Radius' measurement from a freely drawn markups curve with all values set to 0.0. It's the simplest approach I've thought of. There are much fewer conditionals to add in many places, and is hence less confusing. Obviously, the module is modifying an input. I suspect high-level devs may object to this.

Please let me know how you would welcome this approach. If you're comfortable with this, I'll then submit a pull request.

Regards.

chir-set:
Hi @lassoan ,

Could you have a look at this [comment](https://github.com/vmtk/SlicerExtension-VMTK/issues/36#issuecomment-913825752) ?

Thanks.

lassoan:
I started to review the code so that I can give a meaningful answer and found a couple of issues (parameter node not working, not polished GUI, etc.) and did not get to review what the best design would be to handle presence/absence of radius measurement.

Adding an all-zero array of radius measurement is very intrusive and misleading change. Adding an array when you actually not have radius measurements is bad because then how would you know that you actually have radius measurements or not? Adding an array with invalid/irrelevant values does not impact just this module, but all the other modules that use scalars. Setting (borderline) invalid 0 radius values in the array is even worse, because it makes 0 radius values much more frequent and so all modules that use radius arrays would need to be prepared to handle this. I'll finish this up and commit this later today.

Most likely adding a few if/else branches to use radius values if available but not break when they are absent would be much easier than adding a fake radius array and deal with all its consequences in this and many other modules. So, don't add an all-zero array.

chir-set:
> how would you know that you actually have radius measurements or not?

[...]

PR #40, opened 2021-08-29
Merge 'Cross-section analysis' module with 'Centerline metrics'

chir-set:
Hi @lassoan,

Following #36

'Cross-section analysis' core functionalities are hereby added to 'Centerline metrics'.

Namely :

 - orthogonal reslicing to centerline points
 - report slice view orientation in Slicer's coordinate system
 - distance of point from a relative origin.
 
Please consider this patch for merge in 'Centerline metrics'.

Regards.

### DIFF
--- a/CenterlineMetrics/CenterlineMetrics.py
+++ b/CenterlineMetrics/CenterlineMetrics.py
@@ -415,30 +415,6 @@ class CenterlineMetricsWidget(ScriptedLoadableModuleWidget, VTKObservationMixin)
     self.updateSliceViewOrientationMetrics()
     
   def updateSliceViewOrientationMetrics(self):
-    """
-    Please remove comment if commit in trunk.
-    Using hexadecimal for degree character.
-    For obscure reasons, using the degree character triggers :
-    
-    Python 3.6.7 (default, Aug 22 2021, 12:05:08) 
-[GCC Clang 12.0.1] on linux2
->>> 
-Traceback (most recent call last):
-  File "<string>", line 1, in <module>
-  File "/home/user/programs/Slicer/lib/Python/lib/python3.6/imp.py", line 170, in load_source
-    module = _exec(spec, sys.modules[name])
-  File "<frozen importlib._bootstrap>", line 618, in _exec
-  File "<frozen importlib._bootstrap_external>", line 674, in exec_module
-  File "<frozen importlib._bootstrap_external>", line 781, in get_code
-  File "<frozen importlib._bootstrap_external>", line 741, in source_to_code
-  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
-  File "/home/user/programs/Slicer/bin/../lib/Slicer-4.13/qt-scripted-modules/CenterlineMetrics.py", line 389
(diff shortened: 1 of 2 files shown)


chir-set:
One functionality of 'Cross-section analysis' is skipped here.

It concerns orthogonal reslicing along arbitrary markups curves, that do not have radii scalars and are freely drawn.

Adding this functionality to 'Centerline metrics' as it is now would be messy and confusing, both in the UI and in code. Moreover, it's out of scope of a module dealing with centerlines.

I propose that 'Cross-section analysis' be removed from the VMTK group, and moved verbatim to 'Legacy' or 'Utilities'.

What's your take on this ?

lassoan:
Thanks a lot for working on this! 

We already reslice image the image along the curve. We just need to offer the users to choose between reslice modes:
- position only: this is the current behavior
- minimum torsion: this is what the curve-to-world coordinate system provides (uses parallel transport), it can take an additional rotation slider as input to spin the cross-section to an anatomically meaningful orientation
- snap to anatomical axis: automatically spins the slice from the orientation that it got from minimum torsion so that up view up is superior

After this is implemented we can simply remove the `Cross-section analysis` module from the package and we don't have to maintain it anymore.

chir-set:
I'm afraid what you envision is out of reach to me. I just don't have a clue about things like 'direction/normal/binormal vectors' or 'spin the cross-section to an anatomically meaningful orientation'.

Doesn't the reformat widget allow to spin a slice view to any orientation already ?

lassoan:
The tricky part is spinning the extracted slice within the slice view so that the up/down and left/right directions are anatomically meaningful. Without that it is easy to get disoriented when reviewing reformatted slices.

The features that you have implemented are already awesome (I particularly like the fast and robust extraction of cross-section from models), it is perfectly fine that you don't have time to implement everything. I'll get to it in the next few days.

chir-set:
Left/right orientation still needs to be addressed. Can be done manually, rotating 180° around the slice view's Y axis. Automatically ? Doubtful.

chir-set:
Hi @lassoan,

Please review the latest commits when possible.

Thanks.

CenterlineMetrics/CenterlineMetrics.py
```diff
@@ -239,20 +411,62 @@ def updateUIWithMetrics(self, value):
       self.ui.derivedDiameterValueLabel.setText(derivedDiameterStr)
     else:
       self.ui.derivedDiameterValueLabel.setText("")
-
+    # Orientation
+    self.updateSliceViewOrientationMetrics()
+    
+  def updateSliceViewOrientationMetrics(self):
+    """
```
lassoan, line 418: Please remove this comment and instead submit an issue to discuss/resolve this.

lassoan (commented):
Very nice work, thank you. I've just added a trivial comment and it should be good to go.

CenterlineMetrics/CenterlineMetrics.py
chir-set: Done. I'll open a new issue ASAP.

PR #42, opened 2021-09-12
Complete merging 'Centerline metrics' and 'Cross-section analysis'

chir-set:
Hi @lassoan ,

Please consider merging these already [discussed](https://github.com/vmtk/SlicerExtension-VMTK/issues/36) patches : 

d15255a - Simplify CenterlineMetrics module GUI [Andras Lasso]
e6831a1 - Don't add invalid radius scalar to markups curves that don't have radius. [SET]
09671b1 - Allow using an arbitrary markups curve. [SET]
fd4cda9 - Fix calculating distances on arbitrary markups curve directly. [SET]
4b13739 - Reset origin when centerline has changed. [SET]
a7e43a2 - Add MIS model visibility to parameter node. [SET]
41da838 - Reset member variables of logic class on scene close. [SET]
537db57 - Remove 'Cross-section analysis' module. [SET]
1ce4952 - Rename 'Centerline metrics' to 'Cross-section analysis'. [SET]

Thank you.


lassoan:
Thanks a lot. I've tested packaging and on both Slicer-4.11 and Slicer-4.13 and it seems to work well. I know that we made some Slicer core fixes in Slicer-4.13, but even without those, I didn't see any obvious issues with Slicer-4.11.

Unless you find any serious issues on Slicer-4.11, I think we can use this latest version on Slicer-4.11, too.

chir-set:
> Unless you find any serious issues on Slicer-4.11

With 4.11, when using a VMTK markups curve with Radius, the following is noted as soon as a surface is selected : 

```
.....
  File "/home/user/tmp/Slicer/SlicerExtension-VMTK/CrossSectionAnalysis/CrossSectionAnalysis.py", line 886, in updateSliceView
    position = self.getCurvePointPositionAtIndex(value)
  File "/home/user/tmp/Slicer/SlicerExtension-VMTK/CrossSectionAnalysis/CrossSectionAnalysis.py", line 880, in getCurvePointPositionAtIndex
    self.inputCenterlineNode.GetCurvePointsWorld().GetPoint(pointIndex, position)
AttributeError: 'NoneType' object has no attribute 'GetPoint'
```

If we click 'Apply' : 

```
Traceback (most recent call last):
  File "/home/user/tmp/Slicer/SlicerExtension-VMTK/CrossSectionAnalysis/CrossSectionAnalysis.py", line 309, in onApplyButton
    numberOfPoints = self.logic.getNumberOfPoints()
  File "/home/user/tmp/Slicer/SlicerExtension-VMTK/CrossSectionAnalysis/CrossSectionAnalysis.py", line 689, in getNumberOfPoints
    return self.inputCenterlineNode.GetCurvePointsWorld().GetNumberOfPoints()
AttributeError: 'NoneType' object has no attribute 'GetNumberOfPoints'
```

So I think markups curves in 4.11 is not compatible with this version of SlicerVMTK.

The extension manager does show a compatible SlicerVMTK though.

chir-set:
@lassoan 
> With 4.11, when using a VMTK markups curve with Radius, the following is noted as soon as a surface is selected :

I made some other tests with 4.11.

The above errors are reproducible when 4.11 loads a scene saved from 4.13.

No errors are reported when using a scene freshly created with 4.11. It can be saved and reloaded, no error happens.

4.11 users are not expected to work with scenes saved from 4.13. So this could be offered to 4.11 users.

PR #48, opened 2021-11-17
Try cpp cross section

chir-set:
Hi @lassoan ,

I am making this *draft* pull request following your recent suggestion.

I am trying to multi-thread cross-section area calculation in 'Cross-section analysis' module.

My recent wanderings do not give satisfactory results. Memory management has improved but is not yet perfect. And there are random crashes for which I am clueless. I suspect concurrent updates to a shared variable/object, but can't identify one.

Valgrind has not been disrupting, I am probably not using it well (memcheck, callgrind).

I am out of ideas now, and hope you could give me some hints to move on.

There's nothing urgent of course.

Regards.


chir-set:
Hi @lassoan,

I managed to make it work with centerline models without any crash. I believe memory management is good too.

However, there's no way to avoid a crash with centerline curves with the same paradigm. Every call to a function of a curve, or vtkMRMLMarkupsCurveNode::New() is unsuccessful. Slicer and this patch are all built with RelWithDebInfo, and there's no way to step into a curve's function. vtkMRMLMarkupsCurveNode does not seem to be a 'normal' class in C++ code called from Python, at least, that's my perception at this point.

Can you give me some hints for a resolution ?

Thank you.

lassoan:
> Every call to a function of a curve, or vtkMRMLMarkupsCurveNode::New() is unsuccessful

If vtkMRMLMarkupsCurveNode::New() crashes then it means that the software did not compile correctly, the header changed after building the DLLs (e.g., you updated Slicer source code but did not rebuild Slicer), or you use incompatible DLLs (for example, use a DLL built in RelWithDebInfo mode in an application built in debug mode).

> Slicer and this patch are all built with RelWithDebInfo, and there's no way to step into a curve's function.

RelWithDebInfo builds are very hard to debug, as the source code that you see can be very different from the source code that runs (methods, variables may disappear if the compiler can optimize them out), etc.

A clean build from scratch in debug mode should solve all these issues. Note that you need to create a different build folder for each build mode.

chir-set:
> A clean build from scratch in debug mode should solve all these issues.

I made a clean debug buil of Slicer, and a clean debug build of SlicerVMTK. Unfortunately, this did not give better results.

Any call to vtkMRMLMarkupsCurve::New(), ::GetCurveDirectionAtPointIndexWorld or ::GetCurvePointToWorldTransformAtPointIndex in C++ in this context crashes Slicer.
::SafeDownCast() and ::GetCurvePointsWorld() are OK. And there's still no way (crash) to step into a function using lldb.

gdb can't be used after ./Slicer --launch bash :

```
#user@asus0 Slicer-build[0]$ gdb bin/SlicerApp-real 
Fatal Python error: init_import_site: Failed to import the site module
Python runtime state: initialized
Traceback (most recent call last):
  File "/home/arc/src/Slicer-SuperBuild9-Debug/python-install/lib/python3.6/site.py", line 553, in <module>
    main()
  File "/home/arc/src/Slicer-SuperBuild9-Debug/python-install/lib/python3.6/site.py", line 539, in main
    known_paths = addusersitepackages(known_paths)
  File "/home/arc/src/Slicer-SuperBuild9-Debug/python-install/lib/python3.6/site.py", line 282, in addusersitepackages
    user_site = getusersitepackages()
  File "/home/arc/src/Slicer-SuperBuild9-Debug/python-install/lib/python3.6/site.py", line 258, in getusersitepackages
    user_base = getuserbase() # this will also set USER_BASE
  File "/home/arc/src/Slicer-SuperBuild9-Debug/python-install/lib/python3.6/site.py", line 248, in getuserbase
    USER_BASE = get_config_var('userbase')
  File "/home/arc/src/Slicer-SuperBuild9-Debug/python-install/lib/python3.6/sysconfig.py", line 601, in get_config_var
    return get_config_vars().get(name)
  File "/home/arc/src/Slicer-SuperBuild9-Debug/python-install/lib/python3.6/sysconfig.py", line 550, in get_config_vars
    _init_posix(_CONFIG_VARS)
  File "/home/arc/src/Slicer-SuperBuild9-Debug/python-install/lib/python3.6/sysconfig.py", line 421, in _init_posix
    _temp = __import__(name, globals(), locals(), ['build_time_vars'], 0)
ModuleNotFoundError: No module named '_sysconfigdata__linux_x86_64-linux-gnu'
```

This patch will therefore get stalled here. If I don't come to any magic in a few weeks, I'll close or delete the PR if this is possible.

Thanks for your input.

lassoan:
Solution for the sysconfigdata issue is described in this note: https://slicer.readthedocs.io/en/latest/developer_guide/debugging/linuxcpp.html?highlight=debugging#analyze-a-segmentation-fault

If you give step-by-step instructions for reproducing the crash (which Slicer and Slicer-ExtensionVMTK branches to build and what to do after Slicer is started up) then I'll investigate this.

CrossSectionAnalysis/cpp/CrossSectionCompute.cxx
```diff
@@ -0,0 +1,340 @@
...
+    os << indent << "numberOfThreads: " << this->numberOfThreads << "\n";
+    os << indent << "inputCenterlineNode: " << this->inputCenterlineNode << "\n";
+    os << indent << "inputSurfaceNode: " << this->inputSurfaceNode << "\n";
+    os << indent << "inputSegmentID: " << this->inputSegmentID << "\n";
+}
+
+// The surface polydata is constant. Create it once only.
+bool vtkCrossSectionCompute::SetInputSurfaceNode(vtkMRMLNode * inputSurface, const std::string& inputSegmentId)
+{
+  inputSurfaceNode = inputSurface;
```
lassoan, line 43: Use VTK naming conventions: member variables are capitalized and always prefixed with `this->`. This is _very_ important, as not being able to clearly distinguish between member and local variables can easily lead to coding errors, such as crashes.

CrossSectionAnalysis/cpp/CrossSectionCompute.cxx
```diff
@@ -0,0 +1,340 @@
...
+  inputSegmentID = inputSegmentId;
+  if (inputSurfaceNode->IsA("vtkMRMLSegmentationNode"))
```
lassoan, line 45: If `IsA` returns true it does not guaranteed that SafeDownCast will return non-nullptr. Therefore, do not use IsA., but call SafeDownCast and check if the returned value is nullptr.

CrossSectionAnalysis/cpp/CrossSectionCompute.cxx
```diff
@@ -0,0 +1,340 @@
...
+  {     
+    std::cout << "Input segment ID is unknown." << std::endl;
+    return;
+  }
+  /*
+   * Divide the number of centerline points in equal blocks per thread.
+   * The last block will include the residual points also.
+   */
+  const unsigned int numberOfValues = crossSectionAreaArray->GetNumberOfValues();
+  unsigned int residual = numberOfValues % numberOfThreads;
```
lassoan, line 86: There is no need for messing with residual. Division operator or integer values will do integer division. result of `numberOfValues/numberOfThreads` is exactly the same as `(numberOfValues - residual) / numberOfThreads`.

CrossSectionAnalysis/cpp/CrossSectionCompute.cxx
```diff
@@ -0,0 +1,340 @@
...
+    
+    /* 
+     * Give each thread a copy of the centerline.
+     */
+    vtkMRMLNode * inputCenterlineNodeCopy;
+    
+    if (inputCenterlineNode->IsA("vtkMRMLModelNode"))
+    {
+        // Give each thread a copy of each centerline.
+        vtkMRMLModelNode * inputModel = vtkMRMLModelNode::SafeDownCast(inputCenterlineNode);
```
lassoan, line 115: This is not a full copy. For example, the node is not added to the scene, the parent transform is not applied to the node, etc. Creating a full copy would be a quite expensive operation and it would be very difficult to ensure that no events are invoked accidentally from any of the MRML nodes, so I would only pass plain VTK objects to the threads, not MRML nodes.

CrossSectionAnalysis/cpp/CrossSectionCompute.cxx
```diff
@@ -0,0 +1,340 @@
...
+                                  closedSurfacePolyDataCopy,
+                                  bufferArrays[i],
+                                  startPointIndex, endPointIndex));
+  }
+  for (unsigned int i = 0; i < threads.size(); i++)
+  {
+      threads[i].join();
+  }
+  // Update the output table columns.
+  mtx.lock();
```
lassoan, line 148: No threads are running anymore, I don't think a mutex is necessary.

CrossSectionAnalysis/cpp/CrossSectionCompute.cxx
```diff
@@ -0,0 +1,340 @@
...
+    {
+        // Cross-section area is not calculated at last point for a model.
+        vtkMRMLModelNode * inputModel = vtkMRMLModelNode::SafeDownCast(inputCenterlineNode);
+        vtkPoints * modelPoints = inputModel->GetMesh()->GetPoints();
+        const unsigned int numberOfPoints = modelPoints->GetNumberOfPoints();
+        if (endPointIndex == (numberOfPoints - 1))
+        {
+            const unsigned int blockLastPointIndex = bufferArray->GetNumberOfTuples() - 1;
+            const unsigned int blockBeforeLastPointIndex = bufferArray->GetNumberOfTuples() - 1 - 1;
+            bufferArray->SetTuple3(blockLastPointIndex,
```
lassoan, line 211: Have you allocated memory for this tuple? If not, then it may corrupt the memory as you write it. Why don't you use InsertNextTuple3 instead?

CrossSectionAnalysis/cpp/CrossSectionCompute.cxx
```diff
@@ -0,0 +1,340 @@
...
+                                   bufferArray->GetTuple3(blockLastPointIndex)[0],
+                                   bufferArray->GetTuple3(blockBeforeLastPointIndex)[1],
+                                   bufferArray->GetTuple3(blockBeforeLastPointIndex)[2]);
+        }
+    }
+    
+    // The centerline copy is not needed anymore.
+    inputCenterlineNode->Delete();
```
lassoan, line 219: Deleting a node triggers a lots of modifications. If there are observers in the main thread then the application will likely crash. The best would be not to pass nodes to the thread, but if you do then don't modify or delete those nodes in the thread.

CrossSectionAnalysis/cpp/CrossSectionCompute.cxx
```diff
@@ -0,0 +1,340 @@
...
+    double center[3] = {0.0, 0.0, 0.0};
+    double normal[3] = {0.0, 0.0, 0.0};
+    if (inputCenterlineNode->IsA("vtkMRMLModelNode"))
+    {
+        vtkMRMLModelNode * inputModel = vtkMRMLModelNode::SafeDownCast(inputCenterlineNode);
+        vtkPoints * modelPoints = inputModel->GetMesh()->GetPoints();
+        const unsigned int numberOfPoints = modelPoints->GetNumberOfPoints();
+        double centerLocal[3] = {0.0, 0.0, 0.0};
+        modelPoints->GetPoint(pointIndex, centerLocal);
+        inputModel->TransformPointToWorld(centerLocal, center);
```
lassoan, line 256: Transform the model and input surface to world coordinate system once, outside the threads. In the threads just process the resulting VTK objects.

CrossSectionAnalysis/cpp/CrossSectionCompute.cxx
```diff
@@ -0,0 +1,340 @@
...
+            double centerLocalInc[3] = {0.0, 0.0, 0.0};
+            modelPoints->GetPoint(pointIndex + 1, centerLocalInc);
+            inputModel->TransformPointToWorld(centerLocalInc, centerInc);
+            for (unsigned int i = 0; i < 3; i++)
+                normal[i] = centerInc[i] - center[i];
+        }
+        else
+        // Return an empty polydata at the last centerline point
+        {
+            return vtkPolyData::New();
```
lassoan, line 271: Avoid transferring ownership between functions. It is generally much safer and cleaner if you create an object in the same method where you delete that object. In this case, you can create the vtkPolyData instance outside and just pass the pointer as an input argument to the function. The function will update the vtkPolyData content. You can use ShallowCopy (or maybe DeepCopy, but probably that's not needed) to modify an existing polydata with a filter output.

CrossSectionAnalysis/cpp/CrossSectionCompute.h
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * This is the main purpose of this class.
+   * The cross-section area and circular equivalent diameter
+   * columns of the output table are updated in parallel.
+   */
+  void UpdateTable(vtkDoubleArray * crossSectionAreaArray, vtkDoubleArray * ceDiameterArray);
+  
+private:
+  unsigned int numberOfThreads = 1;
+  vtkMRMLNode * inputCenterlineNode = NULL;
```
lassoan, line 63: This is unsafe. The inputCenterlineNode will be deleted as soon as the caller does not need the object anymore. Instead, you can use a vtkSmartPointer<vtkMRMLNode>, which increments the reference counter (preventing deletion) until this vtkCrossSectionCompute class needs the node.

CrossSectionAnalysis/cpp/CrossSectionCompute.h
```diff
@@ -0,0 +1,109 @@
...
+  vtkMRMLNode * inputSurfaceNode = NULL;
+  // Created by ::SetInputSurfaceNode.
+  vtkSmartPointer<vtkPolyData> closedSurfacePolyData = vtkSmartPointer<vtkPolyData>::New();
```
lassoan, line 66: It would be cleaner to do this assignment in the constructor

CrossSectionAnalysis/cpp/CrossSectionCompute.h
```diff
@@ -0,0 +1,109 @@
...
+  std::string inputSegmentID = "";
```
lassoan, line 67: No need to initialize it to empty string (it is already initialized to that).

CrossSectionAnalysis/cpp/CrossSectionCompute.h
```diff
@@ -0,0 +1,109 @@
...
+class vtkCrossSectionComputeWorker
+{
+public:
+    
+    vtkCrossSectionComputeWorker();
+    
+    void operator () (vtkMRMLNode * inputCenterlineNode,
+                                 vtkPolyData * closedSurfacePolyData,
+                                 vtkDoubleArray * bufferArray,
+                                 unsigned int startPointIndex,
```
lassoan, line 80: use vtkIdType instead of int for point indices

lassoan (commented):
I've added a number of comments that might help in revealing memory corruption errors.

CrossSectionAnalysis/cpp/CrossSectionCompute.h
chir-set: Thank you for all your notes, I'll address them one by one.

For this one, the problem is that with 

`vtkStandardNewMacro(vtkCrossSectionCompute);`

I can't declare and define a default constructor  anymore.

I understood, perhaps badly, that because this class derives from vtkObject, instructions like 

```
  static vtkCrossSectionCompute *New();
  vtkTypeMacro(vtkCrossSectionCompute, vtkObject);
  void PrintSelf(ostream& os, vtkIndent indent) override;

  vtkStandardNewMacro(vtkCrossSectionCompute);
```

become compulsory.

CrossSectionAnalysis/cpp/CrossSectionCompute.cxx
chir-set: OK for the 'this->' qualification. 

> member variables are capitalized

You mean, use for example : 

```
vtkSmartPointer<vtkMRMLNode> ANODE;
double AVALUE;
```

instead of 

```
vtkSmartPointer<vtkMRMLNode> aNode;
double AVALUE;
```

lassoan: Member variable name starts with uppercase letter (`this->Something`). Local variable starts with lowercase (`something`). Global constants and enums can be all caps (`SOMETHING`).

CrossSectionAnalysis/cpp/CrossSectionCompute.h
lassoan: Keep using the default New() method. It calls the constructor.

By the way, I don't see your constructor and a virtual destructor. Please add those, if you haven't added them already.

chir-set:
> If you give step-by-step instructions for reproducing the crash

Here is a workflow, and patches required to build Slicer on my machine. I'll make the changes you suggest progressively.

1. Build Slicer at 6c9782419d68, master branch.
2. Build SlicerExtension-VMTK at da5aeafe623b, TryCppCrossSection branch.
3. Copy <Slicer-Extension-VMTK_build_tree>/inner-build/lib/* to <Slicer_build_tree>/lib/
4. Launch Slicer in the build tree.
5. Create a segmentation with model and curve centerlines.
6. Try 'Cross-section analysis' modules with both centerlines

/////////////////////////////////////////////////////////////////////////////////

**Specific to my Arch Linux box**

cmake configurations :

\# Slicer - The tirpc includes are specific to Arch.
```
export CFLAGS="-I/usr/include/tirpc";
export CXXFLAGS="-I/usr/include/tirpc";
cmake -DSlicer_VTK_VERSION_MAJOR:INT=9 -DQt5_DIR:PATH=/usr/lib/cmake/Qt5 -DBUILD_TESTING:BOOL=OFF -DCMAKE_BUILD_TYPE:STRING=Debug -DADDITIONAL_CXX_FLAGS:STRING=-I/usr/include/tirpc -DSlicer_USE_GIT_PROTOCOL:BOOL=OFF ../Slicer
```

\# SlicerExtension-VMTK
```
EXTENSION_BUILD_DIR=`pwd`

export CFLAGS="-I/usr/include/tirpc";
export CXXFLAGS="-I/usr/include/tirpc";
cmake -DSlicer_DIR:PATH=/home/arc/src/Slicer-SuperBuild9-Debug/Slicer-build -DCMAKE_BUILD_TYPE:STRING=Debug -DCMAKE_CXX_FLAGS:STRING="-I/usr/include/tirpc" -DADDITIONAL_CXX_FLAGS:STRING="-I/usr/include/tirpc" -DSlicer_USE_GIT_PROTOCOL:BOOL=OFF -DVMTK_USE_VTK9:BOOL=ON $HOME/src/SlicerExtension-VMTK
```


**\#These patches are needed on Arch.**
/////////////////////////////////////////////////////////////////////////////////
```
commit 27afda0e5e318ed242c54320dd7db42f02bd5958
Author: SET <chir.set@free.fr>
Date:   Mon Nov 22 10:30:48 2021 +0100

    No WERROR

diff --git a/CMakeLists.txt b/CMakeLists.txt
index ab3bcf1f..f95ad1f1 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,7 @@ SET(CMAKE_REQUIRED_INCLUDES)
 SET(CMAKE_REQUIRED_LIBRARIES)
 SET(CMAKE_REQUIRED_FLAGS)
 if (CMAKE_BUILD_TYPE STREQUAL "Debug")
-  OPTION(ENABLE_WERROR "Treat warnings as errors - default is ON for Debug, OFF otherwise." ON)
+  OPTION(ENABLE_WERROR "Treat warnings as errors - default is ON for Debug, OFF otherwise." OFF)
 else ()
   OPTION(ENABLE_WERROR "Treat warnings as errors - default is ON for Debug, OFF otherwise." OFF)
 endif ()
```
/////////////////////////////////////////////////////////////////////////////////
```
commit 544969c2a157b5f31e24439bdb0957677a8bde9a
Author: SET <chir.set@free.fr>
Date:   Mon Nov 22 09:55:12 2021 +0100

    Fix typedef.

diff --git a/check/check.h b/check/check.h
index 0f96f65..d6ccf85 100644
--- a/check/check.h
+++ b/check/check.h
@@ -29,7 +29,7 @@
 #endif
 
 #if defined(HAVE_CC_SHA256_CTX)
-typedef CC_SHA256_CTX lzma_sha256_state;
+typedef SHA256_CTX lzma_sha256_state;
 #elif defined(HAVE_SHA256_CTX)
 typedef SHA256_CTX lzma_sha256_state;
 #elif defined(HAVE_SHA2_CTX)
```
/////////////////////////////////////////////////////////////////////////////////
```
commit 23534b79ab5de9f3e6f53cfa384d275bd24584df
Author: SET <chir.set@free.fr>
Date:   Sat Jul 10 09:26:43 2021 +0200

    Include <limits>

diff --git a/Common/Core/vtkGenericDataArrayLookupHelper.h b/Common/Core/vtkGenericDataArrayLookupHelper.h
index ab9d57248f..a0ae9b9935 100644
--- a/Common/Core/vtkGenericDataArrayLookupHelper.h
+++ b/Common/Core/vtkGenericDataArrayLookupHelper.h
@@ -27,6 +27,7 @@
 #include <cmath>
 #include <unordered_map>
 #include <vector>
+#include <limits>
 
 namespace detail
 {
diff --git a/Common/DataModel/vtkPiecewiseFunction.cxx b/Common/DataModel/vtkPiecewiseFunction.cxx
index 22eca0bc22..4019ce93f7 100644
--- a/Common/DataModel/vtkPiecewiseFunction.cxx
+++ b/Common/DataModel/vtkPiecewiseFunction.cxx
@@ -24,6 +24,7 @@
 #include <iterator>
 #include <set>
 #include <vector>
+#include <limits>
 
 vtkStandardNewMacro(vtkPiecewiseFunction);
 
diff --git a/Filters/HyperTree/vtkHyperTreeGridThreshold.cxx b/Filters/HyperTree/vtkHyperTreeGridThreshold.cxx
index a16bb27fc6..1052192c61 100644
--- a/Filters/HyperTree/vtkHyperTreeGridThreshold.cxx
+++ b/Filters/HyperTree/vtkHyperTreeGridThreshold.cxx
@@ -27,6 +27,7 @@
 #include "vtkHyperTreeGridNonOrientedCursor.h"
 
 #include <cmath>
+#include <limits>
 
 vtkStandardNewMacro(vtkHyperTreeGridThreshold);
 
diff --git a/Rendering/Core/vtkColorTransferFunction.cxx b/Rendering/Core/vtkColorTransferFunction.cxx
index 55c046b4df..2674caaea7 100644
--- a/Rendering/Core/vtkColorTransferFunction.cxx
+++ b/Rendering/Core/vtkColorTransferFunction.cxx
@@ -23,6 +23,7 @@
 #include <iterator>
 #include <set>
 #include <vector>
+#include <limits>
 
 vtkStandardNewMacro(vtkColorTransferFunction);
```
///////////////////////////////////////////////////////////////////////////////// 
```
commit c655fbe74db2510bd58b6073edb688ddd3bc8239
Author: SET <chir.set@free.fr>
Date:   Mon Nov 22 11:11:01 2021 +0100

    Include usual limits; fix LibArchive, LZMA.

diff --git a/Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.cxx b/Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.cxx
index b64a450bd..6996a3a58 100644
--- a/Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.cxx
+++ b/Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.cxx
@@ -13,6 +13,7 @@
 #include <vtkTimerLog.h>
 
 #include "FibHeap.h"
+#include <limits>
 
 vtkStandardNewMacro(vtkImageGrowCutSegment);
 
diff --git a/SuperBuild/External_LZMA.cmake b/SuperBuild/External_LZMA.cmake
index 6e4014f40..de7dd69d6 100644
--- a/SuperBuild/External_LZMA.cmake
+++ b/SuperBuild/External_LZMA.cmake
@@ -35,7 +35,7 @@ if((NOT DEFINED ${proj}_INCLUDE_DIR
   # See https://github.com/python/cpython/blob/3.6/PCbuild/readme.txt
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "v5.2.2"
+    "544969c2a157b5f31e24439bdb0957677a8bde9a"
     QUIET
     )
 
diff --git a/SuperBuild/External_LibArchive.cmake b/SuperBuild/External_LibArchive.cmake
index 5a9e7c14f..80a78fbfb 100644
--- a/SuperBuild/External_LibArchive.cmake
+++ b/SuperBuild/External_LibArchive.cmake
@@ -44,7 +44,7 @@ if((NOT DEFINED LibArchive_INCLUDE_DIR
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "34940ef6ea0b21d77cb501d235164ad88f19d40c"  # master (v3.4.3) with fix-enum-cases
+    "27afda0e5e318ed242c54320dd7db42f02bd5958"  # master (v3.4.3) with fix-enum-cases ADAPTED DEBUG
     QUIET
     )
 
diff --git a/SuperBuild/External_VTK.cmake b/SuperBuild/External_VTK.cmake
index cc095bc40..4b9198b84 100644
--- a/SuperBuild/External_VTK.cmake
+++ b/SuperBuild/External_VTK.cmake
@@ -177,7 +177,7 @@ if((NOT DEFINED VTK_DIR OR NOT DEFINED VTK_SOURCE_DIR) AND NOT Slicer_USE_SYSTEM
     set(_git_tag "97904fdcc7e73446b3131f32eac9fc9781b23c2d") # slicer-v8.2.0-2018-10-02-74d9488523
     set(vtk_egg_info_version "8.2.0")
   elseif("${Slicer_VTK_VERSION_MAJOR}" STREQUAL "9")
-    set(_git_tag "98d686bb509f46543ebe04fdd5d120fdab87893b") # slicer-v9.0.20201111-733234c785-v3
+    set(_git_tag "2a55e1a1ce0e956b8a9ae7024f670e2cc78c233c") # slicer-v9.0.20201111-733234c785-v3 ADAPTED DEBUG
     set(vtk_egg_info_version "9.0.20201111")
   else()
     message(FATAL_ERROR "error: Unsupported Slicer_VTK_VERSION_MAJOR: ${Slicer_VTK_VERSION_MAJOR}")
```

chir-set:
Hi @lassoan ,

I've made a number of changes as per your recommendations. Please have a look when possible.

Please note that it has not been possible to avoid passing MRML nodes to thread workers. There, TransformPointToWorld is needed for centerline models, and GetCurveDirectionAtPointIndexWorld or GetCurvePointToWorldTransformAtPointIndex is needed for centerline curves. Now the thread workers work on independent copies of the input centerline. These copies do not need a scene. So it should not be a problem to pass thread workers MRML node copies, IMHO.

chir-set:
Hi @lassoan,

I've made some changes based on [here](https://github.com/Slicer/Slicer/blob/19d2cbe4cfb5cd3d651f7cdfee1958d1f159d941/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx#L915) and [here](https://github.com/vmtk/SlicerExtension-VMTK/blob/81255c23d77e1549b82f4f21c2af7939282b020a/CrossSectionAnalysis/CrossSectionAnalysis.py#L1028) to calculate cross-section area
in parallel with markups curve. vtkParallelTransportFrame is used to feed
the threads with the curve's polydata and tangents.

Please advise what can be optimized, and if it's getting near for a merge.

Thank you.

lassoan:
I see that some commits have been pushed. Let me know if this PR is ready or you have any specific questions about it.

chir-set:
I think this branch is fit enough for a merge, pending your approval from an expert point of view.

You may also overview another branch, WithWallSurface, when you find some free time. It relies on an [external](https://github.com/chir-set/ExtraMarkups) dependency. Overall, I consider it to be a valuable addition to this module, and for clinicians dealing with vascular diseases.

chir-set:
Hi @lassoan ,

Have you been able to consider this draft pull request ? What can be improved to submit a formal PR? Should it be dropped completely ?

Thanks for considering.

PR #50, opened 2021-12-10
Allow to mirror slice views along X and Y axes.

chir-set:
Hi @lassoan ,

Please consider this patch in branch MirrorXY in my fork.

I noticed that reslicing with SetSliceToRASByNTP almost always inverts left/right and/or anterior/posterior orientation. This is disturbing in orthogonal reslicing with axial navigation, which is the most usual analysis as most centerlines are nearly vertical.

This patch does not modify existing functionalities. It only adds two options.

Thank you.

### DIFF
--- a/CrossSectionAnalysis/CrossSectionAnalysis.py
+++ b/CrossSectionAnalysis/CrossSectionAnalysis.py
@@ -25,7 +25,7 @@ class CrossSectionAnalysis(ScriptedLoadableModule):
     self.parent.contributors = ["SET (Surgeon) (Hobbyist developer)", "Andras Lasso (PerkLab)"]
     self.parent.helpText = """
 This module describes cross-sections along a VMTK centerline model, a VMTK centerline markups curve or an arbitrary markups curve. Documentation is available
-    <a href="https://github.com/vmtk/SlicerExtension-VMTK/tree/master/CrossSectionAnalysis">here</a>.
+    <a href="https://github.com/vmtk/SlicerExtension-VMTK/">here</a>.
 """
     self.parent.acknowledgementText = """
 This file was originally developed by SET. Many thanks to Andras Lasso for sanitizing the code and UI, and for guidance throughout.
--- a/CrossSectionAnalysis/CrossSectionAnalysis.py
+++ b/CrossSectionAnalysis/CrossSectionAnalysis.py
@@ -121,6 +121,8 @@ class CrossSectionAnalysisWidget(ScriptedLoadableModuleWidget, VTKObservationMix
     self.ui.longitudinalSpinSliderWidget.connect("valueChanged(double)", lambda value: self.setValueInParameterNode("LongitudinalSpinAngleDeg", value))
     self.ui.showMISDiameterPushButton.connect("toggled(bool)", lambda value: self.setValueInParameterNode("ShowMISModel", "True" if value else "False"))
     self.ui.showCrossSectionButton.connect("toggled(bool)", lambda value: self.setValueInParameterNode("ShowCrossSection", "True" if value else "False"))
+    self.ui.axialSliceHorizontalFlipCheckBox.connect("clicked()", lambda: self.setValueInParameterNode("AxialSliceHorizontalFlip", str(self.ui.axialSliceHorizontalFlipCheckBox.isChecked())))
+    self.ui.axialSliceVerticalFlipCheckBox.connect("clicked()", lambda : self.setValueInParameterNode("AxialSliceVerticalFlip", str(self.ui.axialSliceVerticalFlipCheckBox.isChecked())))
 
     # connections
     self.ui.applyButton.connect('clicked(bool)', self.onApplyButton)
@@ -140,6 +142,8 @@ class CrossSectionAnalysisWidget(ScriptedLoadableModuleWidget, VTKObservationMix
     self.ui.togglePlotLayoutButton.connect("clicked()", self.togglePlotLayout)
 
     self.ui.outputPlotSeriesTypeComboBox.connect("currentIndexChanged(int)", self.setPlotSeriesType)
+    self.ui.axialSliceHorizontalFlipCheckBox.connect("clicked()", self.setHorizontalFlip)
+    self.ui.axialSliceVerticalFlipCheckBox.connect("clicked()", self.setVerticalFlip)
 
     # Refresh Apply button state
     self.updateGUIFromParameterNode()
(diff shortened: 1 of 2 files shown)


lassoan (commented):
Thank you, this will be useful!
The mirror is basically is almost the same operation as rotations and spins, so it should be implemented exactly the same way and same place as RotationAngleDeg.

Please handle the mirrorx/y variables right after RotationAngleDeg, and use the same kind of connections, callbacks, etc. (and there should be no need for setMirrorXAxis and setMirrorYAxis).

chir-set:
> use the same kind of connections

Please look at the latest patch, it's possible I've not fully understood what you meant.



> same place as RotationAngleDeg
> right after RotationAngleDeg

Do you mean strictly the next lines after those dealing with RotationAngleDeg?

chir-set:
Please see [here](https://disk.yandex.com/d/tD3dpehW6onaRw) an MRB file that might convince of the usefulness of mirroring.

It concerns axial browsing with orthogonal reslicing. Spinning does not restore common IS viewpoint. Arbitrary rotation around the slice view's X/Y axes in this mode is not implemented, and is moot because it would no longer be perpendicular to the centerline. In this case, mirroring around the Y-axis is the solution.

I agree that in most cases, spinning on its own might be sufficient. There are corner cases though.

CrossSectionAnalysis/CrossSectionAnalysis.py
```diff
@@ -121,6 +121,8 @@ def setup(self):
     self.ui.longitudinalSpinSliderWidget.connect("valueChanged(double)", lambda value: self.setValueInParameterNode("LongitudinalSpinAngleDeg", value))
     self.ui.showMISDiameterPushButton.connect("toggled(bool)", lambda value: self.setValueInParameterNode("ShowMISModel", "True" if value else "False"))
     self.ui.showCrossSectionButton.connect("toggled(bool)", lambda value: self.setValueInParameterNode("ShowCrossSection", "True" if value else "False"))
+    self.ui.mirrorXCheckBox.connect("clicked()", lambda: self.setValueInParameterNode("MirrorXAxis", str(self.ui.mirrorXCheckBox.isChecked())))
```
lassoan, line 124: "Mirror" term would be unusual and misleading (as it could indicate that a 3D mirroring is done, which would be a huge issue in a medical image). "Flip" is more accurate, more clear, and used much more commonly.

You don't define X and Y axes directions. Use horizontal and vertical instead. These directions are simpler and well-defined (since they refer to how the image appears on the screen).

These are only applied to the axial slice (not to the longitudinal), so it is better to reflect it in the name.

So, please rename `mirrorXAxis` everywhere to `axialSliceHorizontalFlip` or  maybe `axialSliceHFlip` - and similarly vertical.

CrossSectionAnalysis/Resources/UI/CrossSectionAnalysis.ui
```diff
@@ -528,7 +531,7 @@ The input centerline is expected to be inside the lumen surface.</string>
       <item row="3" column="1">
        <widget class="ctkCollapsibleGroupBox" name="CollapsibleGroupBox_2">
         <property name="title">
-         <string>Spin</string>
+         <string>Spin / Mirror</string>
```
lassoan, line 534: Flip

CrossSectionAnalysis/Resources/UI/CrossSectionAnalysis.ui
```diff
@@ -612,13 +622,41 @@ The input centerline is expected to be inside the lumen surface.</string>
            </property>
           </widget>
          </item>
-         <item row="1" column="0">
-          <widget class="QLabel" name="label_2">
+         <item row="2" column="0">
+          <widget class="QLabel" name="mirrorLabel">
            <property name="text">
-            <string>Long.:</string>
+            <string>Mirror:</string>
```
lassoan, line 628: Mirror -> Flip everywhere

CrossSectionAnalysis/Resources/UI/CrossSectionAnalysis.ui
```diff
@@ -612,13 +622,41 @@ The input centerline is expected to be inside the lumen surface.</string>
...
            </property>
           </widget>
          </item>
+         <item row="2" column="1">
+          <layout class="QHBoxLayout" name="mirrorHorizontalLayout">
+           <item>
+            <widget class="QCheckBox" name="mirrorXCheckBox">
+             <property name="toolTip">
+              <string>Rotate by 180° around the slice view's X-axis.
```
lassoan, line 637: mirrorX would be 180deg rotation around the Y axis. Anyway, use the horizontal and vertical flip terms instead.

lassoan (commented):
Looks good, we should just use more accurate terms. See details inline.

chir-set:
I'll make the changes ASAP, thanks.

chir-set:
@lassoan 

All changes have been made upon your instructions. It should be good now. Thank you.

CrossSectionAnalysis/CrossSectionAnalysis.py
```diff
@@ -121,6 +121,8 @@ def setup(self):
     self.ui.longitudinalSpinSliderWidget.connect("valueChanged(double)", lambda value: self.setValueInParameterNode("LongitudinalSpinAngleDeg", value))
     self.ui.showMISDiameterPushButton.connect("toggled(bool)", lambda value: self.setValueInParameterNode("ShowMISModel", "True" if value else "False"))
     self.ui.showCrossSectionButton.connect("toggled(bool)", lambda value: self.setValueInParameterNode("ShowCrossSection", "True" if value else "False"))
+    self.ui.axialSliceHorizontalFlipCheckBox.connect("clicked()", lambda: self.setValueInParameterNode("HorizontalFlip", str(self.ui.axialSliceHorizontalFlipCheckBox.isChecked())))
```
lassoan, line 124: "AxialSlice" should be added to the parameter name, too.

lassoan (approved):
Great, thanks! I've just added a minor comment, please fix that and then squash all the commits and it should be ready to merge.

chir-set:
Hi @lassoan ,

I was about to 'Squash and merge' and stopped at the last moment because I don't understand why the 'Squash and merge' message includes older commits that are already in the master branch. I certainly don't want to mess up things in the master branch of SlicerVMTK.

For example, c60c9b4e is already committed in the master branch, and many others.

I'm ready to 'Squash and merge' myself, with an adequate message pertaining to the object of this branch, i.e., adding a new flipping functionality. But I wish to understand why the Github 'Squash and merge' function pulls in older commits already in master, and above all, if this implies messing a master branch, which would be obviously terrible.

I understand you are very busy, please take your time.

Regards.

lassoan:
If there are no conflicting changes then "Squash and merge" would do the right thing (you would end up with a single commit, with differences compared to the latest master version).

However, for complex and risky operations like this, it is usually better to do the change locally, verify that everything is good, and then push it to github. To practice, you can create a new branch (so that you can start from scratch if you mess up something). If you are not confident, you can even backup your complete repository folder on your computer so that you can revert your changes.

What you need to do are the followings:
- fetch the changes from upstream (vmtk/SlicerExtension-VMTK) github.com/vmtk/vmtk), this gets all the latest commits to your computer (does not change your branch)
- rebase your branch on the master branch in vmtk/SlicerExtension-VMTK (this rewrites the commit history so that all your changes appear to be done compared to the latest version of the master branch; this is the risky step)
- squash (combine) all your commits that were done after the latest commit on the master branch

If you just push the content of your branch as is then I can do these steps for you, it would not be a problem at all, but it may be a good learning opportunity if you at least try to do it (read up on each step to understand what it does and why it is needed).

To avoid your repository getting in this state (old commits showing up in your pull request), always fetch&rebase your branch before making any modifications to the code.

chir-set:
> it may be a good learning opportunity if you at least try to do it

That's my take indeed, I'll push a targeted commit for the purpose of this branch only. Thank you for the detailed explanation.

chir-set:
I could rebase and squash my local branch on master, resulting in one single commit regrouping all relevant purposed changes. Of course, 'git push origin MirrorXY' fails.

I see 3 options :

1. git push --force origin MirrorXY. It should be harmless as no one is working on this branch. Then merge with vmtk:master within this pull request. Don't know if it would work.

2. Close this pull request, delete the MirrorXY branch in my fork, push the rebased and squashed branch in my fork, create a new pull request and merge. It should work, but will be messy in the pull request history.

3 Clone vmtk:master and patch with the rebased and squashed commit as a *.diff file, commit and push the master branch from my machine. Then close this pull request.

What would be the best solution ?

lassoan:
No need to create a new branch. After you change the commit history, you need to force-push.

PR #51, opened 2021-12-19
Reduce UI simulation

chir-set:
Reduce UI simulation

Adapt to remote changes in SegmentEditorExtra Effects :

 - Use FloodFilling.ROI parameter to set the ROI combobox
 - Use ApplyToAllVisibleSegments of SplitVolume to limit to selected segment

Other local changes to reduce UI simulation :
 - Use functions of
 slicer.modules.SegmentEditorWidget.editor.mrmlSegmentEditorNode()
 to reset masking options
 - Use floodFillFromPoint to initiate flood filling
 - Don't look for stray volume nodes in the subject hierarchy folder created
 by Split Volume, since we can split on a selected segment

Other UI simulations will be addressed later.

### DIFF
--- a/CrossSectionAnalysis/CrossSectionAnalysis.py
+++ b/CrossSectionAnalysis/CrossSectionAnalysis.py
@@ -25,7 +25,7 @@ class CrossSectionAnalysis(ScriptedLoadableModule):
     self.parent.contributors = ["SET (Surgeon) (Hobbyist developer)", "Andras Lasso (PerkLab)"]
     self.parent.helpText = """
 This module describes cross-sections along a VMTK centerline model, a VMTK centerline markups curve or an arbitrary markups curve. Documentation is available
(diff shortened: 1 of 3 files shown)


lassoan:
If you modify the Segment Editor module"s settings then you may interfere with an ongoing segmentation. Instead, it is simpler and safer to create a new segment editor widget in your module. See examples in the script repository, such as https://gist.github.com/lassoan/2d5a5b73645f65a5eb6f8d5f97abf31b

chir-set:
> create a new segment editor widget in your module

I tried it, it works, with one reserve.

The new segment editor widget in its floating window is not really isolated from the common docked widget. Every segmentation and every segment created in one widget appears also in the other one. The best way to avoid interference seems therefore to create a new segmentation completely managed by a module, even in the docked widget.

It is of high interest however if all tasks can be done headless. I know now where to get the relevant code. Thank you.


I'll

PR #52, opened 2022-01-05
Rightly manage an input surface, plus one more fix.

chir-set:
Hi @lassoan ,

Please consider these patches fixing a few things, exposed from [this](https://discourse.slicer.org/t/saving-markups-scalar-values-to-a-file/21287/14?u=chir.set) discussion.

Thank you.


lassoan (approved):
Thank you, looks good.

PR #53, opened 2022-01-13
Stenosis measurement

chir-set:
Hi @lassoan,

Following our discussion [here](https://github.com/Slicer/ExtensionsIndex/pull/1800#issuecomment-1009024258), I submit this pull request accordingly.

The module names have been changed, and their description pages made more verbose and oriented towards stenosis evaluation.

Please let me know what optimizations can be made.

Regards.


StenosisMeasurement2D/StenosisMeasurement2D.py
```diff
@@ -0,0 +1,750 @@
...
+    """
+    Called when the user opens the module the first time and the widget is initialized.
+    """
+    ScriptedLoadableModuleWidget.__init__(self, parent)
+    VTKObservationMixin.__init__(self)  # needed for parameter node observation
+    self.logic = None
+    self._parameterNode = None
+    self._updatingGUIFromParameterNode = False
+    """
+    N.B. - We store slice orientation in NthControlPointOrientationMatrixWorld.
```
lassoan, line 47: Orientation is saved correctly in recent Slicer Preview Releases. This version will be released as the new stable version within a few weeks.

StenosisMeasurement2D/StenosisMeasurement2D.py
```diff
@@ -0,0 +1,750 @@
...
+    Unfortunately, this data is not saved in the markups JSON file,
+    though there is an orientation key.
+    """
+    self.currentFiducialNode = None
+    self.currentControlPointID = "-1" # API gets/sets it as a string
+    self.optionResetControlPointOrientationMatrix = False # Checkable menu item
+    self.optionCreateCutModel = False # Checkable menu item
+    """
+    We are using JumpToPointEvent of the fiducial display node.
+    It gets fired 3 times when we click on a control point.
```
lassoan, line 57: It is fired only once. You can confirm this using this code snippet:

```python
fdn=getNode('F').GetDisplayNode()
observation = fdn.AddObserver(slicer.vtkMRMLMarkupsFiducialDisplayNode.JumpToPointEvent, lambda a, b: print("JumpToPointEvent!"))
```

If you see the event called multiple times then most likely you add the observer multiple times.

StenosisMeasurement2D/StenosisMeasurement2D.py
```diff
@@ -0,0 +1,750 @@
...
+    actionRemoveRow.setData(OUTPUT_TABLE_MENU_REMOVE_ROW)
+    actionRemoveRow.connect("triggered()", self.onTableMenuItem)
+    
+    # Simple menu item to remove all rows.
+    actionEmptyTable = self.tableMenu.addAction("Empty table")
+    actionEmptyTable.setData(OUTPUT_TABLE_MENU_EMPTY_TABLE)
+    actionEmptyTable.connect("triggered()", self.onTableMenuItem)
+    
+    self.tableMenu.addSeparator()
+    # Check menu item to show hide a model of the output polydata.
```
lassoan, line 165: For this it would be simpler to add a vtkBitArray column to the table, which appears as a checkbox. The checkbox is always visible and users can quickly toggle it with a single click.

image

If you add a "Delete selected rows" as a separate button then you can remove the menu completely, making your code simpler.

StenosisMeasurement2D/StenosisMeasurement2D.py
```diff
@@ -0,0 +1,750 @@
...
+    https://discourse.slicer.org/t/observing-a-change-in-sliceoffset-within-a-segmenteditoreffect-class/950/5
+    """
+    # # The slot used by timer cannot have (caller, event) parameters.
+    self.fiducialDisplayNode = None # caller
+    self.fiducialDisplayNodeEvent = None # event
+    self.fiducialDisplayNodeObservation = None
+    self.timer = qt.QTimer()
+    # Menus
+    self.tableMenu = qt.QMenu()
+    self.optionsMenu = qt.QMenu()
```
lassoan, line 68: Is this menu shown when you click on a markup point in a view? For that the clean way would be to register custom actions to the ViewContext menu in a subject hierarchy plugin. See example here: https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#subject-hierarchy-plugin-offering-view-context-menu-action

lassoan (approved):
I did not have time for an in-depth review but looks good overall. I've added some comments that could help to simplify the implementation of the module a little bit.

StenosisMeasurement2D/StenosisMeasurement2D.py
chir-set: > Is this menu shown when you click on a markup point in a view?

No, the menu is shown on right click in the 'Options' collapsible area.
menu_context

StenosisMeasurement2D/StenosisMeasurement2D.py
chir-set: Fixed, it was fired in multiple places in the module's code.

StenosisMeasurement2D/StenosisMeasurement2D.py
chir-set: This one is embarrassing . I found two options.

1. Use a simple boolean value for the column. The table widget shows a checkbox, but it is always left aligned. It's OK functionally, simpler to code, but visually disturbing.

2. Add a custom widget to each cell of the column. This widget has a layout where a checkbox is centered. It's more satisfactory visually. At a functional cost. The checkbox does not trigger anything in the table when toggled, namely, it does not select a row. Connecting a lambda function to the checkbox 'toogled()' signal results in the function always working in the last row. The most acceptable solution is to hide the checkbox by default, connect the table to its 'currentCellChanged()' signal. When the user selects a cell, we know the previous and current cells, and can show/hide the checkbox. It's more complex code, with a visual drawback nevertheless : the checkbox should only appear in the current row. Else, toggling a checkbox in another row does not change current row, and operation occurs in an unwanted object.

I opted for option 2. But can revert to option 1 if you think we should prefer code simplicity namely.

lassoan: I agree that a center alignment would look nicer. That can be fixed in the table view in Slicer core. Could you take a look at it?

StenosisMeasurement2D/StenosisMeasurement2D.py
lassoan: How would users discover that they can bring up a menu by right-clicking there? Why not adding these in a collapsible group ox instead (collapsed by default)? The groupbox would give the hint to the users that they can find additional options there.

chir-set: > How would users discover that they can bring up a menu by right-clicking there?

A tool tip showing "Right click for more" hints for more options. I agree that it's not the *traditional* way in Slicer, I just preferred a lightweight UI. It can of course be changed, please try it once for a personal 'feel' when you have time. I'll change it if you prefer to keep the traditional design.

StenosisMeasurement2D/StenosisMeasurement2D.py
chir-set: > That can be fixed in the table view in Slicer core. Could you take a look at it?

There's perhaps some misunderstanding here. I'm not using a table viewable in the 'Tables' module, nor under the right side views. It's rather a qt.QTableWidget() directly, place in the module area itself. Centering the default checkbox in a cell should be done in Qt itself, AFAIU. There are many internet links where developers complain of this problem. A table cell can align strings only, and nothing else unfortunately.

StenosisMeasurement2D/StenosisMeasurement2D.py
lassoan: Discoverability and consistency is important. Hiding an instruction in a tooltip to bring up a right-click menu is there does not fulfill any of these. 

If you find that the collapsible groupbox is too big then you can hide the additional buttons using the double-arrow button. That is very small, so it won't make the GUI look complex, but it is still recognizable and used elsewhere in Slicer for this.

StenosisMeasurement2D/StenosisMeasurement2D.py
lassoan: I would try to avoid creating a table in Python. It is just too much bookkeeping to make it all work correctly. It is simpler to create a vtkMRMLTableNode and add a MRML table view to see its content. If you set the table column to a vtkBitArray then it appears as a checkbox. Using the  MRML table has the advantage that its content is preserved when you save the scene and its content is also available as a CSV file when you save the scene. What do to show in that table? Can you attach a screenshot?

The checkable item's checkbox is indeed always aligned to the left, which is really, really annoying, because the checkbox works so well other than this issue. It takes a lot of effort to create a checkbox that is centered in a table cell.

chir-set: > What do to show in that table? Can you attach a screenshot?
StenosisMeasurement2D_0

There is no real need to save the table's content in the scene as input parameters and options are saved with the scene. Applying them again is very fast. A MRML table view would upset the UI much, as all slice and 3D views will be resized, and the user will have to select again his preferred layout when switching to another module. This won't resolve the checkbox alignment problem indeed. So it's preferable to keep a QTableWidget in the module area itself.

A vtkMRMLTableNode and a MRML table view are of interest when there are large amount of data from which we plan to plot graphs, like in 'Line profile' or 'Cross-section analysis' modules, or that we may need to export for external statistical analysis. Here, a table is a convenient way to show a result. In practice, this result is rather transient in nature, and does not need long term storage, the more so that one click will produce it again.

StenosisMeasurement2D/StenosisMeasurement2D.py
chir-set: The option menu is now removed. All options are managed with usual widgets.

StenosisMeasurement2D/StenosisMeasurement2D.py
lassoan: The control point, segment, surface area measurements could be useful to be saved automatically as a CSV file but I agree that for the model visibility and segment visibility columns MRML table would not be very useful (because it is just needed for visualization, not that important outside Slicer) and probably it would not simplify much (because you would need to synchronize the vtkBitArray items with associated model and segment visibility).

What would be useful and simplify your code a lot if there was a table for displaying/editing markups control point measurements. Because this table here is just that: a table to display values that are associated with control points. (We call these arrays "measurements" but these are often not computed automatically, but they just store static values.) 

Unfortunately, we don't have the markup control point measurement widget yet, so going with a custom table makes sense.

lassoan:
Thank you for all the updates. I don't have any more comments, this should be good to go.

lassoan:
I would just recommend to squash the commits (have only 1 commit instead of 19).

chir-set:
It's squashed and merged. Thank you for your coaching.

PR #57, opened 2022-03-11
Replace masking option constants to prevent errors after removing dep…

cpinter:
…recated set

Re https://github.com/Slicer/Slicer/pull/6250


lassoan:
Thank you!

PR #61, opened 2022-07-12
Fix/tests

RafaelPalomar:


lassoan (approved):
Sorry, somehow I missed this. Even though some of the changes are just workarounds, they make sense, as it is not useful to report errors on the dashboard. We'll refine the implementation later.

Issue #64, opened 2022-11-11
`Curve centerline extraction` and `Fiducial centerline extraction` modules fail to load


lassoan:
`Curve centerline extraction` and `Fiducial centerline extraction` modules fail to load because of missing dependencies.

Only include modules in `self.parent.dependencies` in the module script if the dependency is in Slicer core, in the same extension as the module, or maybe in an extension that is listed among the extension's dependencies. For all other modules, check if required modules are installed when that module is actually needed (or maybe when the user opens the module) and display a message. For example "This module requires ... extension, please install it using the Extensions manager"; or even offer to the user to [install it automatically](https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#download-and-install-extension).

lassoan:
This pull request that has just been merged (https://github.com/vmtk/SlicerExtension-VMTK/pull/61) removed the dependency but I don't think a user-friendly error message is displayed when the dependency is missing - that would need to be implemented.

chir-set:
> that would need to be implemented.

Ok, I see how to offer installing an extension  or displaying a message if an extension is missing.

Good for the lambda user, bad for geeks who build Slicer.

I usually build Slicer and all extensions I use, and install them all through a bash script. The self built extension packages are just unpacked in the Slicer base directory, *because*, Slicer does not offer a command line option to install/uninstall extension packages. It won't be efficient to install extensions from file packages one by one in the UI every time someone rebuilds Slicer and extensions.

A message for missing extensions may still be displayed in the status bar, it will obviously be less alerting.

So the user's comfort implies complicating the developer's  workflow.

What's your take on this insight ?

lassoan:
> Slicer does not offer a command line option to install/uninstall extension packages

It does - see example in the script repository. You can install from the extensions catalog or you can install from a downloaded and built extension package. If you cannot find an example for something you want to do, or the example does not work, then let me know.

lassoan:
It should be possible to maintain or improve developer convenience independently from the user to xoerience.

chir-set:
I committed bb51f63679 that should nicely fix this issue.

PR #67, opened 2022-11-13
Install extension dependency

chir-set:
QuickArterySegmentation and GuidedArterySegmentation require
    SegmentEditorExtraEffects. Propose to install this extension if
    necessary when a module is shown.
    
    In non-developer mode only.


chir-set:
@lassoan 

I request your review of this patch regarding #64, for any optimization before merging.

PR #82, opened 2023-03-24
Use a markups Shape::Tube node to limit a zone of interest.

chir-set:
The initial zone of interest is a regular tube.

This patch allows to use a markups Shape::Tube node as 'ROI', as a second zone limiting method.

@lassoan


lassoan (approved):
Thanks, it looks good to me.

chir-set:
> Thanks, it looks good to me.

Great, I'm merging it then.

Regards.

PR #85, opened 2023-05-11
With wall surface

chir-set:
Hi @lassoan ,

Could you please consider this pull request ?

It proposes 2 main enhancements : 

- Parallel computation of a segmented artery's lumen cross-section area along its centerline,
- Integrate a markups Shape::Tube node as arterial wall surface for cross-section analysis. The wall surface provides its axial spline as centerline.

Thanks and regards.


chir-set:
No sign of interest, closing.

lassoan:
It is not a lack of interest, it is just that there is no dedicated funding of my time for VMTK, so it is hard to allocate time to review such large pull requests. I trust your judgement - if you are confident that it does not break things then let's integrate this.

chir-set:
> no dedicated funding of my time for VMTK

Ok, I do understand that.

> that it does not break things

It does not break things on Linux. However, I can't tell at all for Windows and MacOS. I don't have the means to check this.

It might therefore be better to leave things as such, until some people can check for these systems. It would be terrible if it is merged and creates problems.

lassoan:
> It does not break things on Linux. However, I can't tell at all for Windows and MacOS

You can merge, wait for the next nightly dashboard build, and if there is a build error then try to fix the error. Usually the fix is quite simple, so it can be done blindly, but if it fails after several attempts then add an issue for it and assign to me.

PR #94, opened 2023-08-13
Add ArterialCalcificationPreProcessor module.

chir-set:
Segment calcifications around an arterial lumen within a margin.


lassoan:
Great, thank you for this contribution!

PR #97, opened 2023-08-30
Use the new ExtensionManagerModel API

chir-set:
Use the new ExtensionManagerModel API to propose installing a dependency extension during setup.

Using [installExtensionFromServer](https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#download-and-install-extension)() referenced in [this](https://discourse.slicer.org/t/install-extensions-from-the-command-line/31439/5) post.


chir-set:
Hi @jcfr,

Could you please check if this patch is the right usage of [installExtensionFromServer](https://github.com/Slicer/Slicer/commit/bda51986c1443c2d53b92a9d4dea4e81e7ac4032)(), fit for a merge ? I understand that it manages testing mode transparently in a non-interactive way.

Thank you.

PR #104, opened 2023-11-28
More centerline derivatives

chir-set:
This PR proposes two modules dealing with centerline models.
 - CenterlineDisassembly: it breaks down a bifurcated centerline model into parts,
 - BifurcationProfiles: it highlights bifurcation splitting lines.
 
Incidentally, the indentation of the list of modules is improved in the main README file.

Please consider for inclusion in SlicerVMTK.

Regards.

PR #105, opened 2023-12-31
Alternate more centerline derivatives

chir-set:
This PR proposes one new module dealing with centerline models, and extends the functionalities of the 'BranchClipper' module.

CenterlineDisassembly: it breaks down a bifurcated centerline model into parts,
BranchClipper: highlight bifurcation splitting lines.

Incidentally,
 - the indentation of the list of modules is improved in the main README file,
 - BranchClipper is improved to prevent a crash when setting array names.

Please consider for inclusion in SlicerVMTK, as an alternative to #104 .

Regards.


chir-set:
@lassoan @pieper  @jcfr

Hi,

I'm considering merging this PR in preference to #104, in 'Rebase and merge' mode. I would prefer an approval from a Slicer main developer before committing, to avoid an unneeded reversal. Please advise.

Thanks and regards.

Docs/CenterlineDisassembly.md
```diff
@@ -0,0 +1,21 @@
...
+
+The components of a centerline can be:
+
+      - bifurcation parts
+      - branches, centerline parts that exclude the bifurcations
+      - centerlines, i.e, one complete centerline from the first endpoint to every other endpoint.
+
+### Disclaimer
+
+Use at your own risks.
```
lassoan, line 21: This is true for the entire 3D Slicer application and all modules in all extensions. If you really want then you can remind people in the extension's documentation. So, I would only add a warning for a specific module if there is something unusual about this module or you have somet more specific information/advice than just "use at your own risks".

lassoan (approved):
It looks good to me, thank you. I've added a few trivial comments inline.

chir-set:
I'll handle all your suggestions in the coming days.

As for the new module template wizard, I suppose you are hinting at the older modules, like Cross-section analysis. I'll update them in the coming weeks as it will take more time. The new parameter node wrapper is indeed more friendly.

Thank your for your comments.

chir-set:
Hi @lassoan,

#105 is merged and SlicerVMTK builds on cdash in all instances except Mac stable. The latter seems to be an already existing problem.

Thank you for your review.

PR #106, opened 2024-01-12
BUG: Fix ExtractCenterline to follow VTK changes

cpinter:
The ExtractCenterline function has been broken since the Slicer commit https://github.com/Slicer/Slicer/commit/d12bcabc2b777512ae310c2117fd2195ab0cccfb (ENH: Update VTK from 9.1.20220125 to 9.2.20230607).

The issue was that vtkThreshold yielded an unstructured grid, the point IDs of which were not in order. To fix the problem, the points in the order which they appear in the output cell are used in the centerline curve.


lassoan (approved):
This looks good, thank you for the fix.

PR #107, opened 2024-01-27
Use the new parameter node wrapper and minor improvements.

chir-set:
Hi @sjh26, @pieper, @lassoan, 

This maintenance PR implements the new parameter node wrapper where it is possible. It is intended to be merged in 'Rebase and merge' mode.

The commits follow the same pattern as discussed [here](https://discourse.slicer.org/t/how-to-handle-the-new-parameter-node-wrapper/33957). Basically, reference the parameter node in the logic in addition to the widget class, and store output objects as well as members of the parameter node class.

Two minor changes are included:
- fix the minimum number of points required by a curve in 'GuidedArterySegmentation',
- rename a widget in 'BranchClipper' removing the 'FooBar' template string.

I request the approval of at least one Slicer developer before merging it.

Thanks and regards.

### DIFF
--- a/BranchClipper/Resources/UI/qSlicerBranchClipperWidget.ui
+++ b/BranchClipper/Resources/UI/qSlicerBranchClipperWidget.ui
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ui version="4.0">
- <class>qSlicerBranchClipperFooBarWidget</class>
- <widget class="QWidget" name="qSlicerBranchClipperFooBarWidget">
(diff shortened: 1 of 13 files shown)


lassoan (approved):
It looks good to me. It seems that you were able to remove quite a bit of code.

chir-set:
@lassoan 

#107 is merged with master.

When the parameter node wrapper becomes compatible with more widgets, I'll update 'CrossSectionAnalysis', and if you wish and if it's necessary, 'ExtractCenterline'.

Thanks and regards.

PR #119, opened 2024-07-29
Handle holes in lumen.

chir-set:
Ignore holes, usually small defects, in a segmented lumen or in a model using vtkPolyDataConnectivityFilter::SetExtractionModeToLargestRegion().


chir-set:
Hi @lassoan,

I hereby request your opinion on this commit.

Holes in a segment may invert the expected result of cross-section analysis. The
surface is calculated from polydata extracted with vtkConnectivityFilter::SetExtractionModeToClosestPointRegion where the reference point is on the centerline.

A nearby hole might be extracted instead of the lumen as shown below.

Current

Proposed


This patch ignores the holes by working on a reworked polydata: the whole lumen polydata is trimmed with vtkPolyDataConnectivityFilter::SetExtractionModeToLargestRegion(), and this result is then passed on for further computations.

I wish to know if you can think of any downside of this handling. I don't readily find any for now.

Thanks and regards.

lassoan:
This would help in most cases, but keeping largest connected region is a blunt tool, with several limitations:
- it would not remove those holes that are connected to the vessel lumen
- the largest connected region is not necessarily the one that is closest to the centerline curve
- the user has no control over what holes to fill (in general, if any automated processing is performed on any data, the user should be able to inspect the result of the automatic processing and if it is not perfect then the should be able to fix it)

Instead of silently running largest connected region extraction, we could run use largest connected region extraction to detect if small islands may be present. If there is significant difference between the full mesh and the largest component then a warning could be displayed to the user and the user could choose to: cancel and fix manually, fix automatically (by Smoothing effect or maybe largest connected region extraction); and after that the user would be offered to verify the lumen (it would be a nice touch if the user could jump to prev/next suspicious cross-section).

chir-set:
> This would help in most cases, but keeping largest connected region is a blunt tool, with several limitations:
> 

I see what you mean, thanks.

I finally think there should only exist an information widget that describes the number of connected regions in the input mesh.

Theoretically, there should be only one connected region. Holes are additional regions, but additional regions may not be holes in the lumen. They may lie outside of the lumen and can also be larger than the lumen.

I'll devise an information widget that should not take much space, the user being free to take more steps on the input.

Thanks for your input.

chir-set:
@lassoan 

I have implemented:
 - identification of regions,
 - browsing to individual regions,
 - activating the 'Paint' effect in-place for a segmentation lumen.
 
Fixing is therefore manual, for segmentations only. Automatic repair of poor segmentations will be error prone, difficult to code and remains probably the user's responsibility.

Waiting for your review, nothing urgent of course.

PR #127, opened 2024-09-05
Indicate ITK and VTK as superbuild dependencies for VMTK

dzenanz:


lassoan (commented):
Thank you, this looks good to me. To make sure that this section is not removed during refactoring in the future, please add a comment to the source code explaining why it is needed (i.e., to allow building as an extension bundled with Slicer).

lassoan (approved):
Thank you.

PR #129, opened 2024-09-06
Allow building extension when bundled with Slicer

dzenanz:


CMakeLists.txt
```diff
@@ -27,8 +27,13 @@ mark_as_superbuild(Slicer_DIR)
 find_package(Git REQUIRED)
 mark_as_superbuild(GIT_EXECUTABLE)
 
-find_package(ExtraMarkups REQUIRED)
-mark_as_superbuild(ExtraMarkups_DIR)
+option(SlicerVMTK_enable_ExtraMarkups_and_StenosisMeasurement3D
```
lassoan, line 30: This should be sufficient (if `ExtraMarkups` is disabled we could then disable all modules that depend on it).
```suggestion
option(SlicerVMTK_USE_ExtraMarkups
```

lassoan (commented):
Added a minor comment, otherwise looks good.

CMakeLists.txt
```diff
@@ -27,8 +27,13 @@ mark_as_superbuild(Slicer_DIR)
 find_package(Git REQUIRED)
 mark_as_superbuild(GIT_EXECUTABLE)
 
-find_package(ExtraMarkups REQUIRED)
-mark_as_superbuild(ExtraMarkups_DIR)
+option(SlicerVMTK_USE_ExtraMarkups
+  "This needs to be disabled when this extension is bundled with Slicer" ON)
```
lassoan, line 31:
```suggestion
  "This needs to be disabled if ExtraMarkups Slicer extension is unavailable" ON)
```

lassoan:
Thank you, merged. Note that we have many extensions with complex dependencies (for example, SlicerIGT, SlicerOpenIGTLink, SlicerIGSIO), and bundling them all in a custom Slicer application works fine, as long as all dependencies between them are specified.

chir-set:
Does Slicer build with this setup? This [include](https://github.com/vmtk/SlicerExtension-VMTK/blob/ebc84e66c8d04dbef6029d6a17584c77ba1ce2dc/CrossSectionAnalysis/Logic/vtkCrossSectionCompute.cxx#L22) might cause trouble.

lassoan:
Good catch @chir-set - indeed the extension build failed when built with SlicerVMTK_USE_ExtraMarkups=OFF (I don't know how it succeeded for @dzenanz). I've pushed a fix: https://github.com/vmtk/SlicerExtension-VMTK/commit/18ee741be988ed59772e776152498c25680d054f

dzenanz:
Ah, probably the reason that it built for me was also declaring ExtraMarkups as a bundled extension:
```cmake
# ExtraMarkups is a dependency of VMTK
set(extension_name "ExtraMarkups")
set(${extension_name}_SOURCE_DIR "${CMAKE_BINARY_DIR}/${extension_name}")
FetchContent_Populate(${extension_name}
  SOURCE_DIR     ${${extension_name}_SOURCE_DIR}
  GIT_REPOSITORY ${EP_GIT_PROTOCOL}://github.com/chir-set/SlicerExtraMarkups.git
  GIT_TAG        a51f6f97b679a9797aa1917ea77b61f2a79b7716
  GIT_PROGRESS   1
  QUIET
  )
message(STATUS "Remote - ${extension_name} [OK]")
list(APPEND Slicer_EXTENSION_SOURCE_DIRS ${${extension_name}_SOURCE_DIR})

# SlicerExtension-VMTK
set(extension_name "SlicerVMTK")
set(${extension_name}_SOURCE_DIR "${CMAKE_BINARY_DIR}/${extension_name}")
set(SlicerVMTK_USE_ExtraMarkups OFF)
mark_as_superbuild(SlicerVMTK_USE_ExtraMarkups)
FetchContent_Populate(${extension_name}
  SOURCE_DIR     ${${extension_name}_SOURCE_DIR}
  GIT_REPOSITORY ${EP_GIT_PROTOCOL}://github.com/VMTK/SlicerExtension-VMTK.git
  GIT_TAG        ebc84e66c8d04dbef6029d6a17584c77ba1ce2dc
  GIT_PROGRESS   1
  QUIET
  )
message(STATUS "Remote - ${extension_name} [OK]")
list(APPEND Slicer_EXTENSION_SOURCE_DIRS ${${extension_name}_SOURCE_DIR})
```
I guess that this is no longer needed with your fix Andaras?

chir-set:
Theroretically, _CrossSectionAnalysis_ could be kept in the embedded suite IF _vtkCrossSectionCompute.cxx_ is modified with conditionals _#if_ or _#ifdef_ that would exclude _vtkMRMLMarkupsShapeNode_. It remains to be known which pre-processor definition to use.

In _CrossSectionAnalysis.py_, _vtkMRMLMarkupsShapeNode_ is always checked before calling its interface, so it should work simply without ExtraMarkups.

As an additional note, _GuidedArterySegmentation_ module should be modified also because _vtkMRMLMarkupsShapeNode_ is [called](https://github.com/vmtk/SlicerExtension-VMTK/blob/18ee741be988ed59772e776152498c25680d054f/GuidedArterySegmentation/GuidedArterySegmentation.py#L168) without any check. Now the function will never be called without ExtraMarkups because the combobox trigger, which is an optional input, is populated with _vtkMRMLMarkupsShapeNode_ only.

PR #131, opened 2024-10-11
Cross-section analysis object management

chir-set:
1. Add a parameter node MRML combo box.
    
    This simplifies the management of multiple surfaces and multiple centerlines.

2. Rework the output plot chart and series nodes.
    
    Formerly, plot chart series nodes were made available in a combobox of the
    module's widget. For one centerline/surface pair, multiple plot series can be
    produced and rendered in a single plot chart node. This patch exposes the plot
    chart node in the module's widget instead of the plot series. The latter can be
    managed in the 'Plots' module and in the plot widget's controller. All plot
    series that can be created for one centerline/surface pair are created.


chir-set:
Hello @lassoan,

Please consider this PR having the following rationale.

Initially, the module was working with the 'Radius' array of a centerline model
or a centerline markups node. There was only one array and its was logical to
expose a plot series node as output.

Then the cross-section area of the lumen with the consequent CE diameter. Two
more arrays were appeared.

Ultimately, the Shape::Tube node joined in, bringing 4 more arrays.

It became hard to handle all these in a study. This PR allows to better organise
a work session. It's much more efficient to conduct an analysis with a
centerline/lumen pair grouped in a parameter set. Each parameter set generates
its own table and it own plot chart node. All plot series of a pair are computed
and the chart node is exposed instead of the individual series nodes.

Testing in multiple scenarios has been conclusive. The change is relatively
'big'. I wish to have your opinion before committing.

Thanks and regards.

CrossSectionAnalysis/CrossSectionAnalysis.py
```diff
@@ -1810,10 +1757,10 @@ def test_CrossSectionAnalysis1(self):
...
 DIAMETER_STENOSIS_ARRAY_NAME = _("Stenosis by diameter (CE)")
 
-MIS_DIAMETER = "MIS_DIAMETER"
-CE_DIAMETER = "CE_DIAMETER"
-CROSS_SECTION_AREA = "CROSS_SECTION_AREA"
-WALL_CE_DIAMETER = "WALL_CE_DIAMETER"
-WALL_CROSS_SECTION_AREA = "WALL_CROSS_SECTION_AREA"
-DIAMETER_STENOSIS = "DIAMETER_STENOSIS"
-SURFACE_AREA_STENOSIS = "SURFACE_AREA_STENOSIS"
+MIS_DIAMETER_PLOT_SERIES_COLOR = (0.0, 0.6, 1.0)
```
lassoan, line 1760: There are several chart color tables that contain colors that are supposed to work well for charts. It would be nicer to use those.

```
rgba=[0,0,0,0]
getNode('DarkBrightChartColors').GetColor(0, rgba)
print(rgba)
```

CrossSectionAnalysis/CrossSectionAnalysis.py
```diff
@@ -1263,32 +1215,54 @@ def updateOutputTable(self, inputCenterline, outputTable):
...
+    plotSeriesNode.SetAndObserveTableNodeID(self.outputTableNode.GetID())
+    # Make sure the plot is in the chart
+    if not self.outputPlotChartNode.HasPlotSeriesNodeID(plotSeriesNode.GetID()):
+      self.outputPlotChartNode.AddAndObservePlotSeriesNodeID(plotSeriesNode.GetID())
+    return plotSeriesNode
+
+  def updatePlot(self, outputPlotChart, outputTable, name=None):
+    # Recreate all output plot series nodes.
+    self.outputPlotSeriesNodes.clear()
+    self.outputPlotChartNode.RemoveAllPlotSeriesNodeIDs()
```
lassoan, line 1238: We should also delete the series nodes from the scene, otherwise we would leave behind a lot of obsolete series nodes each time we update the plot. Of course, we don't want to accidentally delete series that some other modules generated and were added to the same chart, so the module could to set some attribute on the series node that it creates and only delete those series that it generated.

CrossSectionAnalysis/CrossSectionAnalysis.py
```diff
@@ -1263,32 +1215,54 @@ def updateOutputTable(self, inputCenterline, outputTable):
...
+    if self.isCenterlineRadiusAvailable():
+      misDiameterSeriesNode = self._createOutputPlotSeries(MIS_DIAMETER_ARRAY_NAME, MIS_DIAMETER_ARRAY_NAME, lengthUnit, MIS_DIAMETER_PLOT_SERIES_COLOR)
+      # self.outputPlotChartNode.SetYAxisTitle(_("Diameter ({unitOfLength})").format(unitOfLength=lengthUnit))
+      self.outputPlotSeriesNodes.append(misDiameterSeriesNode)
+
+    if self.lumenSurfaceNode:
+      ceDiameterSeriesNode = self._createOutputPlotSeries(CE_DIAMETER_ARRAY_NAME, CE_DIAMETER_ARRAY_NAME, lengthUnit, CE_DIAMETER_PLOT_SERIES_COLOR)
+      self.outputPlotSeriesNodes.append(ceDiameterSeriesNode)
+      crossSectionAreaSeriesNode = self._createOutputPlotSeries(CROSS_SECTION_AREA_ARRAY_NAME, CROSS_SECTION_AREA_ARRAY_NAME, areaUnit, CROSS_SECTION_AREA_PLOT_SERIES_COLOR)
+      self.outputPlotSeriesNodes.append(crossSectionAreaSeriesNode)
```
lassoan, line 1254: We should not display length and area values in the same chart. What would be the meaning of the vertical axis then?

lassoan (commented):
It looks good to me, I've just added a few comments inline.
I haven't tried the module, but if we create a single chart and we create several series with different units (mm, cm2, cm3) then what would be the meaning of the chart? What would the Y axis be?

CrossSectionAnalysis/CrossSectionAnalysis.py
chir-set: The nodes are now removed from scene.

Also:
Using a distinct logic instance for each MRML scripted node. They are tracked and applied when a scripted node is selected.

CrossSectionAnalysis/CrossSectionAnalysis.py
chir-set: Well the units are appended to the plot series name, and this appears in the legend. The Y axis shows a range that varies with the selected series.

plot_all

plot_single

lassoan: Diameter, cross-section, and percentage should never appear in the same plot, as they are different quantities. VTK might support dual axis charts, which could be used to display absolute and percentage value; or length and area quantities in one chart. Can you check if VTK indeed supports it? (there is a method to get left, right, top, bottom axis, but I'm not sure if you can set different ranges for them)

chir-set: > Can you check if VTK indeed supports it?
I could not find this support. In any case, only dual axis won't suffice.

> should never appear in the same plot
In that case, the plot rework becomes pointless. I'll limit this PR to the parameter set addition. It will still be easier to work with multiple centerlines.

chir-set:
Hi @lassoan ,

Please check the last patch that a plot chart node to a single series. Thanks.

lassoan:
I'm not sure when I will have time for reviewing and testing the code. If you could give a short summary instead that could help in giving you feedback faster. Thank you!

chir-set:
> If you could give a short summary 

I'll make a few more tweaks and write a comprehensive list of the changes.

chir-set:
Hello @lassoan ,

Here is a resume of the (last)final changes:

>     Use multiple parameter nodes:
>         - generated in a combobox handling vtkMRMLScriptedModuleNode nodes
>         - intended to group a centerline/surface input pair: one pair, one parameter
>         node
>         - group outputs: one table and one chart for one parameter node
>         - restore pre-computed results when a parameter node is selected
> 
>     Improve handling of the plot series:
>         - do not enforce the name of a plot series node
>         - synchronize the plot chart with the parameter node
>         - keep the display of a single series in a chart
> 
>     Remove unused variables, functions and comments.

Waiting for your comment.

Regards.

lassoan (approved):
Thank you, the changes look good to me.

chir-set:
Thank you for your review, it's merged. I'll next investigate if a parameter set node may fit some other modules.

PR #132, opened 2024-10-25
Add a parameter set controller

chir-set:
GuidedVeinSegmentation

Using a single wrapped parameter node and multiple vtkMRMLScriptedModuleNode instances, each defining a parameter set.

At the same time, use the wrapped parameter node in logic.

### DIFF
--- a/GuidedVeinSegmentation/GuidedVeinSegmentation.py
+++ b/GuidedVeinSegmentation/GuidedVeinSegmentation.py
@@ -72,10 +72,8 @@ class GuidedVeinSegmentationWidget(ScriptedLoadableModuleWidget, VTKObservationM
         ScriptedLoadableModuleWidget.__init__(self, parent)
         VTKObservationMixin.__init__(self)  # needed for parameter node observation
         self.logic = None
-        self._wrappedParameterNode = None
-        self._mrmlParameterNode = None
+        self._parameterNode = None
         self._parameterNodeGuiTag = None
-        self._parameterNodeBeingModified = False
 
     def setup(self) -> None:
         """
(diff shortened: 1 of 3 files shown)


chir-set:
Hello @lassoan,

I request you opinion about this patch.

The goal is to add a parameter set node to most modules in SlicerVMTK where the
parameter node wrapper is being used.

The chosen method is as follows:
 - generate a single wrapped parameter node from the logic class
 - connect it once only
 - never replace this parameter node
 - reset it to default values when the scene is closed
 - generate vtkMRMLScriptedModuleNode instances from an MRML combobox
 - initialise the MRML parameter node with default values
 - when an MRML parameter node becomes active, update the wrapped parameter node
 accordingly (this updates the UI widgets)
 - when a UI widget is modified, update the active MRML parameter node.

This is not a formal pull request yet, I wish to read your thoughts about this.
It's the simplest effective approach I have found when the parameter node
wrapper is used in a module. The other solution would be to drop the wrapper,
but it's not where Slicer is going as I understand.

Thanks and regards.

lassoan:
Thanks for working on this! It would be greate to use parameter node in VMTK modules.

I would recommend to start from the latest version of the scripted module template: create a module by the same name as the existing module and copy over all the relevant code from the old version of the module. If you find anything wrong with the template we should fix the template.

chir-set:
Thank you for your suggestion. After comparing the current module template (created a new module) and the GuidedVeinSegmentation module (which was created using that template at that time, i.e, with the parameter node wrapper), I consider that the module is conform to the current template.

The problem seems to be that a wrapped parameter node is never disconnected from the UI. As we add more wrapped parameter nodes, any UI interaction is propagated to all of them.

This code snippet illustrates this.

```
import GuidedVeinSegmentation
mainWindow().moduleSelector().selectModule('GuidedVeinSegmentation')

logic = getModuleLogic("GuidedVeinSegmentation")
widget = getModuleWidget("GuidedVeinSegmentation")

parameterNode = logic.getParameterNode() # A wrapped parameter node.

print("Logic generated parameter node:", parameterNode.shellMargin) # 18.0

scriptedModuleNode = slicer.vtkMRMLScriptedModuleNode()
additionalParameterNode = GuidedVeinSegmentation.GuidedVeinSegmentationParameterNode(scriptedModuleNode) # A wrapped parameter node.
widget.setParameterNode(additionalParameterNode)

print("Additional parameter node:", additionalParameterNode.shellMargin) # 18.0

print("---------", "Additional parameter node is current", "---------")
print("Setting value in additionalParameterNode")
additionalParameterNode.shellMargin = 16.5 # Or UI interaction
print("Logic generated parameter node:", parameterNode.shellMargin) # 16.5
print("Additional parameter node:", additionalParameterNode.shellMargin) # 16.5
print("         ---")
print("Setting value in parameterNode")
parameterNode.shellMargin = 10.2 # Or UI interaction
print("Logic generated parameter node:", parameterNode.shellMargin) # 10.2
print("Additional parameter node:", additionalParameterNode.shellMargin) # 10.2

print("---------", "Logic generated parameter node is current", "---------")
widget.setParameterNode(parameterNode)

print("Setting value in additionalParameterNode")
additionalParameterNode.shellMargin = 22.8 # Or UI interaction
print("Logic generated parameter node:", parameterNode.shellMargin) # 22.8
print("Additional parameter node:", additionalParameterNode.shellMargin) # 22.8
print("         ---")
print("Setting value in parameterNode")
parameterNode.shellMargin = 19.7 # Or UI interaction
print("Logic generated parameter node:", parameterNode.shellMargin) # 19.7
print("Additional parameter node:", additionalParameterNode.shellMargin) # 19.7

```

lassoan:
Please report this to the Slicer Forum - Development category, @mention Sam Horvath and Kyle Sunderland. We need to follow the template and if there are any issues or concerns then the template should be fixed.

chir-set:
This PR is dropped, #143 being a cleaner method to use parameter sets.

PR #134, opened 2024-11-05
BranchClipper: Allow using a model surface input.

chir-set:
BranchClipper

 - use a model surface input in addition to a segmentation
 - create segments from an input segmentation
 - create models from an input model
 - do not track branch outputs.

### DIFF
--- a/BranchClipper/Resources/UI/qSlicerBranchClipperModuleWidget.ui
+++ b/BranchClipper/Resources/UI/qSlicerBranchClipperModuleWidget.ui
@@ -15,9 +15,9 @@
   </property>
   <layout class="QVBoxLayout" name="verticalLayout">
    <item>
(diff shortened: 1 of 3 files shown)


chir-set:
Hello @lassoan ,

Please review this patch that basically allows processing a model surface in addition to a segmentation. This is long overdue, I don't remember why I missed it. At the same time, the output branches are no longer tracked, i.e, they are created again if the user applies repeatedly.

Thanks and regards.

lassoan (approved):
Thank you, it looks good. I've just added a note inline about potential naming improvements.

chir-set:
The widgets are now in a collapsible button with 'Inputs' instead of 'Parameters'.
The surface label is now 'Tree tube'. The input surface combobox is updated to 'Pick an input segmentation or model node'.

It's now merged, thank you.

PR #138, opened 2024-12-06
Add EditCenterline module.

chir-set:
This module pre-fits a Shape::Tube markups node along an input markups curve that represents the axis of a non-aneurysmal artery.


chir-set:
Hello @lassoan,

I am hereby proposing this module for inclusion in SlicerVMTK.

Its main purpose is to simplify the drawing of a Shape::Tube markups node
around an artery. The main input is a single markups curve.

It uses GuidedVeinSegmentation to segment the lumen of an artery, diseased or
not. The resulting segment is chopped at each control point of the input markups
node, and a pair of points of the Shape::Tube node is added at the farthest
points of each contour.

As the name implies, 'Pre-fit', a post-processing step remains necessary to
verify and reposition the control points of the tube.

Overall, I found it to be of significant help on real cases.

If you find it to be a worthy addition here, we may next discuss renaming the
module if you deem it necessary.

Thanks and regards.

lassoan:
Thank you for working on this. I'm trying to understand what is the difference compared to `Extract Centerline` module, which already computes maximum inscribed sphere position and radius for each curve point. Instead of adding a new module, wouldn't it be sufficient to make `Extract Centerline` module accept a tube markup output? When a tube was specified as input the control points of the tube would be set up based on the radius (+/- radius distance from the centerline along the curve normal direction).

chir-set:
A. The expected workflow if ExtractCenterline is modified would be:

 - segment an artery, diseased or not
 - extract a centerline curve
 - either resample the curve to much fewer control points
 - or define in the to-be-written code a reasonable interval between the curve's control points
   - specify this interval in the UI
 - generate the tube  based on the MIS radius
 - fix control point position.
 
B. The workflow in this proposal:

 - place an arbitrary curve
 - choose a dimension profile
 - generate the tube
 - fix control point position.
 
The implications with A:
 - create a dependency of ExtractCenterline on ExtraMarkups¹ (recall the SlicerVMTK_USE_ExtraMarkups option)
 - complicate the code, UI and maintenance of ExtractCenterline²
 - complicate the workflow.

The implications with B:
 - leave everything as such
 - add something new, simple, independent and hence discardable.

¹ is not something desirable, ² neither in my view.

The pictures below show that both A and B can be done

PreFitTube_ArbitraryCurve

PreFitTube_ExtractCenterlineCurve

The notion of sane arteries handled in test data, and diseased arteries handled daily is of capital importance. The purpose of the tube is to represent the arterial wall, it is not visible in any imaging modality and can only be estimated by the observer. The tube is not meant to represent the lumen, it is easily segmented, and segmented (nearly) exactly.

Thank you for your review.

lassoan:
OK, if the dimension profile does not need a centerline with radius information (so that you can use it with a hand-drawn centerline) then it makes sense to have this as a separate module.

The actual goal of the centerline seems to be manually drawing/editing centerline and radius; the end result could be in the same format as extract centerline results; and the module could be usable for editing automatically extracted or manually drawn centerline result. Therefore, I would recommend to consider naming the module `Edit centerline` and allow it to specify centerline + radius; and in the long term also allow importing/exporting of radii values from/to markup curves and models.

chir-set:
Ok, there are more functionalities to add.

However, I'm soon away for 2 weeks and somehow busy. I'll look into that in January.

lassoan:
Sounds good! These features don't have to be added now, I just tried to find out what could be the long-term scope of the module because that helps with finding a module name that is self-explaining to users, and that remains accurate as the module is further developed. "Pre-fit" sounded like a first step of a task but it is better to implement a complete task in a module.

chir-set:
Hello @lassoan,

I renamed the module to EditCenterline.

ExtraMarkups has been updated to show the spline of a Tube shape, this has meaning in this module.

Later on, I plan to use an input centerline model/curve. The main question to resolve is the number of control point pairs to use in the output Tube node.

Thank you for considering.

chir-set:
> Later on, I plan to use an input centerline model/curve.

This is implemented in c4e3395b79.

lassoan:
I agree, adding a radius scale could be useful sometimes. If it is not too difficult to implement it and you can add it to the GUI so that it does not make the GUI more complicated (e.g., added in an advanced section, collapsed by default) then it could be good.

chir-set:
@lassoan, would you mind if I make `ExtractCenterline::_addCurveMeasurementArray` public in this PR? This would avoid code duplication here.

chir-set:
Hello @lassoan ,

I think this PR should be ready for a merge. If you agree, I can do that.

Please let me know.

chir-set:
Hello @lassoan ,

I went forward and merged EditCenterline module. I found it really helpful and saves a lot of time.

Regards.

PR #143, opened 2025-03-10
Add a parameter set selector.

chir-set:
General pattern:

 - do not use the parameter node wrapper
 - generate vtkMRMLScriptedModuleNode instances from the parameter set selector
 - let each widget update the parameter node
 - update widgets with a usual 'updateGUIFromParameterNode()' function


chir-set:
Hello @lassoan,

I wish to add a parameter set to most modules of SlicerVMTK since it is a very convenient facility.

I am pushing this modification of 'ArterialCalcificationPreProcessor' as a completed sample. I can proceed with the other modules according to the same pattern if you are agree with this pattern.

Slicer has a hard bearing on the 'parameter node wrapper', it must nevertheless be put aside when using a parameter set. It seems the wrapper does not expect that its inner parameter node might be replaced.

Looking forward to hear from you on this topic.

Thanks and regards.

lassoan:
> Slicer has a hard bearing on the 'parameter node wrapper', it must nevertheless be put aside when using a parameter set.

The [parameter node is a singleton by default (one instance per scene)](https://github.com/Slicer/Slicer/blob/432c725ea2748b3ab8404ab18d4724cb6ba0f654/Base/Python/slicer/ScriptedLoadableModule.py#L308), but you can set `self.isSingletonParameterNode=False` in your logic's constructor to allow the user to choose between many parameter set nodes. See for example [here](https://github.com/Slicer/Slicer/blob/432c725ea2748b3ab8404ab18d4724cb6ba0f654/Modules/Scripted/SegmentStatistics/SegmentStatistics.py#L389C14-L389C22).

I agree that the parameter node wrapper has the disadvantage that it makes things a bit more opaque (things happen under the hood that are not that easily accessible or readily understandable), but it simplifies code and reduces the amount of code that developers need to maintain. However, the Slicer core team has no capacity to support multiple scripted module patterns and the decision was made to go with the parameter node wrapper (the alternative would have been to scrap the entire parameter node wrapper infrastructure and it felt that would not have been the right direction). If you find any issues with the parameter node wrapper or something is unclear then please let me or other Slicer developers know about it and we'll try to make things easier (fix bugs, improve documentation, examples, etc.).

You can always decide to do things your own way, and for example choose not use parameter node wrapper, but then it would become much more difficult to help you, because I would need to spend time with understanding your new custom pattern, how it works, what the various advantages/disadvantages, etc. It is much easier to justify the time investment into improving the common pattern that we recommend to everyone to follow.

chir-set:
Thank you for your reply.

The 'isSingletonParameterNode' attribute is new to me; it does not however concern the parameter node wrapper and won't solve its inability to be selected in a qMRMLNodeComboBox since it is python based. I already [introduced troubles](https://github.com/Slicer/Slicer/issues/8009#issuecomment-2439726904) with the wrapper and follow-up stopped at some point.

Anyway, the pattern I am speaking of has nothing exotic, it's the same one used in 'ExtractCenterline' for example. A qMRMLNodeComboBox provides instances of vtkMRMLScriptedModuleNode, each one independent of the other. The module's widget uses the selected parameter node and updates the GUI with a dedicated function. Widgets in-turn update the parameter node when modified. This has also the advantage of handling all widgets, because the parameter node wrapper does not handle all widgets yet. It's easy to understand constraints about time and funding. The fact is that at some point, even without the need of a parameter set, we have to use both the wrapper and the pre-wrapper pattern for not handled widgets.

I consider that a 'parameter set' can be a powerful option, it took me some time to realise this. Even if it implies reverting to the pre-wrapper approach, it seems worthwhile to me.

lassoan:
When you switch the parameter set node, the parameter node wrapper keeps working, automatically updating the GUI from the newly selected parameter set node.

Indeed a few widgets do not have GUI connectors yet. Additional GUI connectors can be defined in any module to cover any missing widget classes, but if you use a Slicer core widget or CTK widget then it makes sense to add the GUI connector to the Slicer core.

lassoan:
Defining parameters and node references; showing/editing them in the GUI; saving/loading then in the scene are all essential features that should not overwhelm developers. We would be interested to hear about your experience and get suggestions about how do you think these could be simplified. Thank you!

chir-set:
> how do you think these could be simplified.

I do not have a clear reply to your question. I guess the parameter node wrapper is intended to address this question. It's limitations drove me to abandon the wrapper when a parameter set is of interest. Perhaps a wrapper as a vtkMRMLNode implemented in C++ would have resulted in efficient concurrent usage, but perhaps only.

This branch has a clean implementation on a traditional proven pattern that does not face any limitation. It works and will survive time. There's more coding obviously, but 'overwhelming' is probably not the right description. Here, a pragmatic approach is being preferred over a conceptual one.

If you agree, I can commit this branch after a few days of testing on real world cases. I won't do it without your consent.

Thanks.

chir-set:
Hello @lassoan,

After reconsidering, I changed my mind and am planning to merge this PR in a few days *unless* you disapprove of it.

After using it for some days, I consider it as a good enhancement. It will not degrade user experience in any way.

Regards.

PR #145, opened 2025-04-12
Fix surface area calculation

chir-set:
1. Fix computation of surface area and more.

The cross-section surface areas were not calculated for model and shape nodes.

At the same time:
 - prefer vtkPolyDataConnectivityFilter to vtkConnectivityFilter
 - stop updating the output table if parallel computing should fail
 - update an erroneous message
 - inform if a cross-section contains holes or branches (although the segment
 does not).

2. Adapt to cmake 4.0.

 - The minimum required version is 3.5.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13.4)
+cmake_minimum_required(VERSION 3.5)
 
 project(SlicerVMTK)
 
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13.4)
+cmake_minimum_required(VERSION 3.5)
```
jamesobutler, line 1: @chir-set The CMake minimum for 3D Slicer is 3.16.3. Where 3.13.4 was actually an old minimum for 3D Slicer. It was updated in https://github.com/Slicer/Slicer/commit/664fecaf8d70acdfaba30a473fe35931a4999e47. You will want to maintain this CMake minimum policy like Slicer.

chir-set: It's updated in #146 that I plan to merge after some more testing.

Thank you.

PR #148, opened 2025-04-30
Clip the lumen within a tube

chir-set:
Clip a lumen inside a tube and more.

StenosisMeasurement3D

 - Add a function to clip a common enclosed surface or none.
 - Add a function to improve the meshing of a polydata.
 - Ensure the scene of the logic class is not null.
 - Clip the lumen inside the tube.
 - Display results according to application-wide unit formats.
 - Tentative normalisation of calculated values: consider the lesion volume and
 stenosis degree per cm.
 - Move the boundary points to the positions referenced in a table row: a simple
 helper function with a single purpose.
 - Rename and make some functions public:
   - Clip() -> ClipClosedSurface()
   - ClipClosed() -> ClipClosedSurfaceWithClosedOutput()
 - Check for invalid normal and points.

CrossSectionAnalysis

  - Allow to clip a lumen if a tube is used as centerline. The lumen can be a
  segment or a model. The created clipped lumen can be used to constrain
  cross-sections within the tube. This improves the reliability of measurements
  at bifurcations.
  - When the UI is manually updated to a paremeter set, don't log an error if
  the input lumen is a model; just don't update the UI.
  - Use the tryWithErrorDisplay() facility on apply.
  - When selecting an input lumen node as a model, don't log and return silently.
  - Fix multi-threaded surface area determination: vtkMassProperties was not
  being updated, resulting in slightly lower values.


chir-set:
Hello @lassaon,

May you overview this PR when possible?

It's _main_ purpose is to clip the lumen enclosed in a tube so as to restrict analysis in that part of the lumen.

Thank you.

PR #150, opened 2025-05-02
Revert the addition of a parameter node in loadable StenosisMeasurement3D

chir-set:
The module does not build for Windows on cdash. It builds for Mac, and for Linux on my machine.

At the same time, call Update() on vtkMassProperties in CrossSectionAnalysis.


lassoan:
Instead of reverting, you can apply these changes: https://github.com/lassoan/SlicerExtension-VMTK/commit/c0a2d8c00121aafb83627e93b24e59fadddfb0b7

chir-set:
The scope of the parameter node has changed in #153. It is now limited to the widget class and is not used in the logic (apart from registering). Just reverting will imply much more rework and I have pending branches after that. I think the core of #153 should be left as such.

PR #153, opened 2025-05-09
Add a parameter set selector.

chir-set:
StenosisMeasurement3D

Using a vtkMRMLScriptedNode instead of a specialised parameter node. Its scope is limited to the widget and is not used in the logic.

At the same time:
 - Updating the lumen mesh in the widget: make it optional.


lassoan:
It is interesting that you use the `vtkMRMLScriptedModuleNode` for loadable modules. This class was implemented as a simple workaround for allowing scripted modules have parameter nodes (as scripted modules cannot create new VTK classes).

It should not be necessary to use this workaround in C++ module. It is very easy to create a new MRML node in C++ to stores various properties and it provides a nice, clean, documented, type-safe interface in both C++ and Python. See for example Crop Volumes module parameter node [header](https://github.com/Slicer/Slicer/blob/main/Modules/Loadable/CropVolume/MRML/vtkMRMLCropVolumeParametersNode.h) and [implementation](https://github.com/Slicer/Slicer/blob/main/Modules/Loadable/CropVolume/MRML/vtkMRMLCropVolumeParametersNode.cxx).

What discouraged you from implementing a new class for your parameter node?

chir-set:
> This class was implemented as a simple workaround for allowing scripted modules have parameter nodes

That's how I understood things.

My initial attempt was with a dedicated class, based on the Crop volume implementation. It worked fine but did not build for Windows, see #150.

I still have the files and can update things to not use vtkMRMLScriptedModule node, but I won't be able to guarantee anything for Windows.

On the other hand, a parameter node does nothing more than providing accessor methods. The vtkMRMLScriptedModule does the same thing with less hassle.

I'll update the module.

lassoan:
OK, I see. I tested that commit and I did not see an issue on Windows and then realized that I would have needed to enable `ExtraMarkups`, etc. and ran out of time.

I've had a look at it now and the build error was justified and very easy to fix:

https://github.com/lassoan/SlicerExtension-VMTK/commit/c0a2d8c00121aafb83627e93b24e59fadddfb0b7

I don't know how it was possible that the build succeeded on any platform without these changes.
Note that explicit adding of include libraries (that you attempted to do later) is not necessary (if targets are set up correctly then adding a library should automatically add the associated include directories).

chir-set:
I put back the MRML parameter node and followed the changes you highlighted above.

I'll wait until tomorrow for any red light from you.

Thanks for having had a look.

PR #169, opened 2025-09-09
Decimate a tube polydata and more.

chir-set:
Improve the title of the plot series.
CrossSectionAnalysis

\- Specify the surface being studied: lumen or wall
\- Use strings defined in the global variables
\- Rename some variables to imply the surface type: lumen or wall
\----------------------------
Inform about empty cross-sections.
CrossSectionAnalysis

If empty cross-sections are generated (lumen or wall), inform it as a status
message that hints at the application log for details.
\----------------------------
Allow optional decimation of the wall polydata.
StenosisMeasurement3D
CrossSectionAnalysis

The purpose of decimating a polydata here is twofold:
 - minimise the causes of failure in vtkBooleanOperationPolyDataFilter
 - minimise the creation of empty cross-sections.

vtkBooleanOperationPolyDataFilter is used to clip the lumen within the tube in
StenosisMeasurement3D, and to clip a lumen model within the tube in
CrossSectionAnalysis. (A lumen segment in the latter module is clipped within the
tube using the segment editor.)

vtkQuadricDecimation is used and its main parameters can be tuned (API only).
The default parameters target minimal decimation.

For cross-sections of the Tube in CrossSectionAnalysis, decimation is tuned on
the higher end of the target reduction ratio to obtain a valid cross-section
polydata (not empty) on the whole spline.

Decimation is always optional. It should be used if any processing fails (lumen
clipping, cross-section). Even then, success is not guaranteed. The processing
may even succeed without any decimation and fails if any pre-processing is
applied.

At the same time:
 - rename some identifiers for clarity in CrossSectionAnalysis mainly,
 - make vtkCrossSectionCompute work on input polydata objects only, instead of
   MRML nodes,
 - restore the display node properties in
   vtkSlicerStenosisMeasurement3DLogic::UpdateSegmentBySmoothClosing(),
 - decrease logging verbosity in CrossSectionAnalysis,
 - decrease the minimal width of the CrossSectionAnalysis widget if there is no
   lumen input,
 - clear the tooltip of the result labels (CrossSectionAnalysis),
 - update inline comments.
\----------------------------

Simple test:

Applying CrossSectionAnalysis on this sample tube as single input shows the effect
of decimation of the tube's polydata (UI toggle button).

```
points = [
  [21.76709747314453, -6.943074315468127, -141.15455578520258],
  [21.76709747314453, 0.7340565618670549, -142.70243884370367],
  [22.93106460571289, -5.7411364332880135, -130.81582541997471],
  [22.93106460571289, 2.284702660247242, -132.18252663080654],
  [24.970205307006836, -5.408451214975221, -121.71147094274264],
  [24.970205307006836, 3.758911148256164, -118.6135870406558],
  [25.731433868408203, -7.779374062090739, -112.82542442116518],
  [25.731433868408203, -1.0793152460757651, -110.38270546164732],
  [25.84928321838379, -5.015497530907004, -101.6005741919926],
  [25.84928321838379, -8.91319337943235, -102.36344558339805]
]

tube = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsShapeNode")
tube.SetShapeName(slicer.vtkMRMLMarkupsShapeNode.Tube)
tube.SetResolution(45)
tube.SetSplineNewInterpolationInterval(True)

for point in points:
  tube.AddControlPointWorld(point)

```

CrossSectionAnalysis_Tube_0

CrossSectionAnalysis_Tube_1


lassoan:
Thanks for the updates. Just a note on Boolean mesh operations: vtkBooleanOperationPolyDataFilter is very unreliable. It often gives incorrect output due completely valid inputs. I would not use it. Instead, try the Combine Models module in Sandbox extension - it is much more robust. It uses a VTK filter that you can use in your module directly.

chir-set:
> try the Combine Models module in Sandbox extension

I indeed saw `vtkPolyDataBooleanFilter` in `Combine Models` and its excellent results ended up in this approach of boolean operation on polydata.

The result with `vtkBooleanOperationPolyDataFilter` is clearly less satisfactory.

If `Combine Models` should be referenced from this module, it implies that the `Sandbox` extension becomes a dependency for `SlicerVMTK` during the build on factory machines. That stopped me from going after `vtkPolyDataBooleanFilter`.

Another option is to integrate the `vtkbool` project files in `StenosisMeasurement3D` with due acknowledgement, like you did in `Combine Models`.

There is one last option that won't involve boolean operations on polydata:
 - since the lumen here is a segment (it has never been considered as a model in the module),
 - and since the result of converting the surrounding tube to a binary label map is very close to the closed surface representation (the tube won't ever be smaller than the image data spacing and it represents the arterial wall, i.e,  no stenosis),

I have a mockup code that can intersect both label maps predictably. The closed surface representation of the resulting label map can then be used faithfully.

This last method is used in `CrossSectionAnalysis` now, by using the `Logical Operations` effect of the segment editor.

Looking forward to read your comments further.

Thank you for having considered.

chir-set:
I'll see what I can come up to, the last option that you mentioned seems less intrusive. I fear that when Slicer starts, the module does not load at all if `libvtkSlicerCombineModelsModuleLogic.so` is not found. If it works, `vtkBooleanOperationPolyDataFilter` should just be dropped.

What do you think about deploying the `vtkbool` minimal implementation files in a sub-directory of the Logic directory?

lassoan:
> I fear that when Slicer starts, the module does not load at all if libvtkSlicerCombineModelsModuleLogic.so is not found

There are no build-time dependencies like in C++, so you don't have to worry about this. If Sandbox extension is not installed then the `import vtkSlicerCombineModelsModuleLogicPython as vtkbool` will throw an `ImportError`. If you put it in a `try`/`except` block then you can easily handle the error.

chir-set:
> import vtkSlicerCombineModelsModuleLogicPython as vtkbool

[vtkBooleanOperationPolyDataFilter](https://github.com/vmtk/SlicerExtension-VMTK/blob/6e787f8d7b6871c1bc450abca1283c3e36985e0c/StenosisMeasurement3D/Logic/vtkSlicerStenosisMeasurement3DLogic.cxx#L661) is being used in `StenosisMeasurement3D` which is a loadable module.

The logic of `StenosisMeasurement3D` is also called from `CrossSectionAnalysis` which is indeed a scripted module. `vtkSlicerCombineModelsModuleLogicPython` can be rightly called here.

The optimal solution would be to implement the `vtkPolyDataBooleanFilter` statically in `StenosisMeasuerement3D`, as a drop-in replacement of `vtkBooleanOperationPolyDataFilter`, and call the logic from `CrossSectionAnalysis`, and of course, from `StenosisMeasurement3D` itself.

chir-set:
Following the [discussion](https://discourse.vtk.org/t/building-vtkbool-as-a-remote-module-does-not-seem-to-complete/16043) on the VTK forum, the optimal solution would be to add `vtkbool` to `vtkAddon`. It would then be available in Slicer at large. I have no idea how to do that.

lassoan:
Yes, unfortunately, VTK developers keep pushing back on making the life easier for VTK user/developer community. vtkAddon and vtkbool are two separate projects, so we cannot add vtkbool into vtkAddon, but we could add vtkbool to Slicer similarly to vtkAddon. I'll ask @jcfr about this at [today's Slicer developer weekly meeting](https://discourse.slicer.org/t/2025-09-16-weekly-meeting/44488). You are very welcome to join this meeting @chir-set (starts in about 15 minutes).

PR #186, opened 2026-08-10
Add per-point manual adjustment of clip plane origin and normal

lassoan:
We got feedback from a clinician that it would be preferable if the plane normal was set automatically when the clipping position is modified. To enable this feature while also allowing manual adjustments, two new buttons were introduced next to "Finish plane editing" button:

image

- enableManualPlaneOrigin (crosshair icon): when pressed, the plane origin is adjusted manually and dragged clip points snap to the visible surface; when released, the clip point is snapped onto the centerline.
- enableManualPlaneNormal (rotation icon): when pressed, the plane can be rotated manually; when released, the plane normal follows the centerline direction at the clip position.

A pressed button stores the current origin/normal as a per-point entry in the ManualClipPlaneOrigins/ManualClipPlaneNormals lists; releasing it removes the entry and snaps the plane back to the centerline. The button states are updated from these lists whenever a clip point is selected, and the buttons are only enabled while a plane is edited.

The existing "Snap clip points to centerline" checkbox (SnapClipPointsToCenterline parameter) keeps controlling the default drag behavior of clip points that have no manual origin override.

Interaction handles that would fight the centerline-based behavior are hidden: in-plane (x/y) translation handles while the origin follows the centerline, and rotation handles while the normal does. The blue rotation ring around the plane normal and the free in-plane rotation ring are always hidden, as rotating the plane about its own normal never changes the cut.

Also fixes a bug where the clip plane appeared at the world origin when first shown: SetNormalWorld called inside a StartModify batch resets the plane's not-yet-synchronized origin, so the origin is now restored explicitly afterwards.


lassoan:
@aabrown100-git FYI, this is a small feature based on Dr Whitehead's feedback.

PR #188, opened 2026-08-31
Add optional face labeling of the clipped surface

aabrown100-git:
Adds a **Label mesh faces** option to Clip Vessel that tags every cell of the output with an integer face id in a cell array (`ModelFaceID` by default — the name SimVascular and its meshing tools read a model's faces by). That is what a CFD setup needs to assign a boundary condition per face, and what lets a remesher keep the wall/cap boundaries as sharp feature edges instead of rounding them off.

The output is coloured by face id automatically, with a legend naming each face after the clip point that made it.

## Id layout

The wall takes id 1 and the cap made at the Nth clip point takes N+2.

Caps are numbered by **clip point**, not by the capping filter's own boundary order. That order comes from `vtkvmtkPolyDataBoundaryExtractor` and shifts as clip points are moved, which would silently renumber the faces of an already-configured simulation between two runs. Each cap is matched to its clip point by the distance from the cap centroid to the ray leaving the clip plane along its normal, so the match survives a flow extension pushing the cap several radii down the removed branch.

If the input surface already carries face labels they are kept, not discarded: compacted onto 1..K in ascending order of their original id, with the wall and then the caps following. A hole that no clip point accounts for is a defect in the mesh rather than a vessel end — a single missing triangle leaves a three-point boundary — so its fill joins the face around it instead of becoming a face of its own.

## Filter behaviour worked around

Three things in the surrounding filters bite here, each found by a wrong result rather than an error, so they are worth calling out:

- **`vtkvmtkCapPolyData` inherits an existing array of the name it is given** and still numbers the new caps from `CellEntityIdOffset` upwards, so a new cap lands on an id already in use and the two faces fuse silently. The caps are therefore routed through a private array name and the user-facing array is built afterwards.
- **`vtkPolyData` indexes its cells verts → lines → polys → strips** while the capping filter copies only the polys. A single vert or line cell — `vtkCleanPolyData` makes them out of degenerate triangles — would shift every label one cell out of step and scatter them across the surface. Only the poly cells' labels are taken.
- **Preprocessing rebuilds the mesh and what it leaves of a cell array cannot be relied on**: decimation drops it, and with a degenerate triangle in the input `vtkTriangleFilter` leaves it at twice the cell count and misaligned, so the labels arrive scrambled rather than missing and no length check downstream can catch it. The labels are re-derived from the original surface by position instead, which also means they now survive decimation.

Applying with preprocessing set to decimate a labeled input asks for confirmation first, since the face boundaries can then shift by about a cell. Only the explicit Apply click warns, so a modal dialog cannot interrupt an interactive plane drag.

## Testing

The module self-test covers the id layout with and without flow extensions, an uncapped surface, and an input that already carries labels.

Beyond that, `ClipVessel/Testing/Python/FaceLabeling/` holds developer regression suites — 209 checks over ten scripts, run with `run_all.sh`. They are deliberately not registered with CTest: most need a Slicer application with the module loaded and assert on MRML display state rather than logic alone. They pin the three filter hazards above, the legend label path (which fails as `(none)` unless `LabelFormat` carries a string specifier alongside `UseColorNamesForLabels`), and the full pipeline on a real vessel surface. Happy to drop them from the PR if you would rather they lived outside the repository.

Two diagnostics ship alongside them for inspecting a real case in a running Slicer: one reports where each face id came from and measures against a cell locator whether the labels stayed on the geometry they started on, the other counts open boundaries per stage to show where an unexpected hole is introduced.


lassoan (approved):
Thank you, nice work!

Identifying boundaries (openings) by re-running the boundary extractor and hoping that it always gives the same order was the current practice in VMTK, but it was somewhat fragile, so I've added a [new filter to VMTK that stores the boundary ID inside the mesh](https://github.com/vmtk/vmtk/pull/505) (one point array stores the boundary ID, and another array the order within that boundary and updated the flow extender and capping filter to use it.

I've also replaced the .sh file for running tests (that only works on macOS) with more standard way of running the tests, which ensures that all the tests are automatically executed in automated builds and submitted to the dashboard; and make it easy to run them using the SlicerPythonTestRunner extension. Instructions are provided in the new Developers.md page.

image

PR #189, opened 2026-09-01
Add a choice of capping method to Clip Vessel

aabrown100-git:
## Summary

Clip Vessel closes the clipped ends with `vtkvmtkCapPolyData`, a fan of triangles from the centre of the hole. That is one of three capping filters VMTK ships and Slicer builds; this exposes the other two in a **Cap method** list next to the cap checkbox:

Center point (default) - `vtkvmtkCapPolyData` - triangle fan to the boundary barycentre:

image

Simple - `vtkvmtkSimpleCapPolyData` - one flat polygon per hole, adds no points:

image

Smooth - `vtkvmtkSmoothCapPolyData` - concentric rings of cells, optionally domed:

image

image

The fan gives a badly shaped element where every triangle meets, and it gets no finer as the boundary does; the ring-based smooth cap meshes a large or elongated boundary far better.

**Center point stays the default**, so existing scenes are unaffected. The smooth capper's two shape settings appear only when that method is selected. *Cap roundness* defaults to 0, which keeps the cap in the plane of the cut exactly as the other two methods do — so switching to smooth changes only how the cap is meshed, not where it sits — and raising it domes the cap outward along the direction the wall was already heading. *Cap rings* (default 8) sets how finely it is meshed. Those are the only two shape parameters the filter has.

The two `vmtksurfacecapper` methods left out do not apply here: `annular` bridges *pairs* of boundaries on a walled surface and needs an even number of them, and `concaveannular` lives in `vtkvmtkcontrib`, which Slicer does not build.

## Details

Three things had to be squared away to slot the new filters in.

**Triangulation.** Neither new capper triangulates what it makes — simple fills a boundary with a single polygon, smooth with rings of quads — and the smooth one reads its input as triangles. The surface is therefore triangulated before capping, placed ahead of the existing face ids being read so they still line up cell for cell, and again afterwards.

**Cap winding.** Both new cappers wind their cap cells the opposite way round from the vessel wall. `vtkvmtkCapPolyData` emits `(p_j, centre, p_{j+1})`, while the other two follow the boundary loop, which is the reverse. Measured on a test tube: wall 100% outward, caps 0%. The caps render as though lit from inside.

**Stale normals.** `vtkvmtkSimpleCapPolyData` is the only one of the three that calls `PassData` on its input's point data (the equivalent line in `vtkvmtkCapPolyData` is commented out). Because it closes a boundary without adding any point, every vertex of its cap is a boundary vertex carrying the normal the vessel wall left there — none of them face along the cap. The cap is then shaded as though it were wall, which looks like streaky shadowing across a flat face.

The last two are handled by a new `orientSurfaceOutwards()` applied after capping: it re-winds the cells outward and drops the normals, leaving the renderer to shade by geometry as it already did for the centre point capper. One wrinkle worth noting for review — `vtkPolyDataNormals` skips its consistency pass entirely when both compute flags are off, so it computes cell normals, re-winds, and the array is dropped again. Cell count, cell order and point count are all preserved, so the face labeling still lines up.

## Testing

The module self test now runs all four configurations (centre point, simple, smooth at roundness 0, smooth at roundness 1) and checks each output is watertight, fully triangulated, consistently outward facing, free of stale normals, and carries one face id per cap — plus that only a smooth cap with a nonzero roundness reaches out of the cut planes.

The filters were also exercised directly against a VMTK build on an open tube:

```
CENTERPOINT    cells=  160 tri=True open=0 ids=[0,1,2] reorientIsNoOp=True staleNormals=False vol=+563.164 diag=21.7256
SIMPLE         cells=  156 tri=True open=0 ids=[0,1,2] reorientIsNoOp=True staleNormals=False vol=+563.164 diag=21.7256
SMOOTH r=0     cells= 1276 tri=True open=0 ids=[0,1,2] reorientIsNoOp=True staleNormals=False vol=+563.164 diag=21.7256
SMOOTH r=1     cells= 1276 tri=True open=0 ids=[0,1,2] reorientIsNoOp=True staleNormals=False vol=+637.300 diag=25.4608
```

Positive signed volume and "re-orienting finds nothing to re-wind" both confirm outward-facing caps. Smooth at roundness 0 reproduces the flat caps' bounding box and enclosed volume exactly.

## Notes for reviewers

- This branch is stacked on `Add optional face labeling of the clipped surface`, which is not yet on `master`, so that commit appears in this diff as well. Happy to rebase if it lands first or if you would rather see the two separately.
- `Docs/ClipVessel.md` gains a *Capping the clipped ends* section.


lassoan (approved):
Thank you, I've rebased on latest master, switched to the new, more robust boundary ID storage (in point data) and fixed a triangulation bug in the smooth capper. It is good to go. I'll push a uniform cap remesh option soon.

PR #195, opened 2026-09-04
Add junction angle quantification to Centerline disassembly

aabrown100-git:
Adds a fourth component to the Centerline disassembly module, **Junction angles**, which measures the angles between the branches that meet at each bifurcation of a centerline. The other three components are unchanged, and no other module is touched.

### The measurement

It is the one VMTK already provides, so this is mostly a matter of wiring the existing filters into the module and presenting the result. For every bifurcation the module computes the bifurcation reference system and the bifurcation vectors with `vtkvmtkCenterlineBifurcationReferenceSystems` and `vtkvmtkCenterlineBifurcationVectors`, the way the `vmtkbifurcationreferencesystems` and `vmtkbifurcationvectors` scripts do in the [geometric analysis tutorial](https://github.com/vmtk/vmtk.github.com/blob/master/tutorials/GeometricAnalysis.md). VMTK stores each vector along the flow direction, so the parent's points towards the bifurcation; the module reverses it, which makes every direction point away from the bifurcation and the angle of a pair simply the angle between the two directions, in the 0-180 degrees range.

Two properties of this definition are worth noting, and are why it was preferred to sampling tangents at a chosen distance along the centerline:

- the vectors are computed on the branches, where each branch keeps its own path through the bifurcation, not on a merged centerline where the branches share a trunk near it;
- the distance over which a direction is measured follows the local maximum inscribed sphere radius, so there is no sampling distance to choose and the measurement behaves the same in large and small vessels.

On synthetic tubes whose axes meet at a known angle the measurement recovers that angle within about one degree, for vessel radii between 2 and 8 mm and independently of the branch lengths.

### What the component creates

- a table with one row per pair of branches per bifurcation: the angle, the angle projected onto the bifurcation plane, the out of plane angle of each branch, the `GroupId` and the parent or child role of both branches, the junction position and the junction degree;
- a curve for each measured direction, in a 'vectors' folder, so that the distance over which each direction was measured can be seen;
- an angle markup for each pair, in 'Child-child angles' and 'Parent-child angles' folders so that either group can be shown or hidden at once.

Rows and nodes carry the `GroupId`s as node attributes, which match the ones on the branch and bifurcation components, so a row can be traced back to a component. Since it always creates its own outputs, the Models and Curves options do not apply to it.

Junction angles

### Tests

Four tests are added to the module's test case, and the existing ones still pass: a bifurcation of known angle, a multifurcation, the table's contents against the annotations, and an end to end run on a Y shaped tube surface, from segmentation through centerline extraction to the angles.

`Docs/CenterlineDisassembly.md` documents the component, the definition of the vectors, and every column of the table.


chir-set:
Hello,

I have not gone through the details of the code and here are my comments.

1. These angles can be valuable to both researchers and clinicians, so the endeavour is fully legitimate.

2. Your patch plugs into the 'CenterlineDisassembly' module while not requiring the purpose of centerline splitting, which defines the module itself.

3. Following ~2, it could well be plugged into other modules also, line 'ExtractCenterline'. This would require just one checkbox as an option when a bifurcation centerline model is created.

4. Hence, I think it would be better to spin off a stand-alone module for your proposal. This would be less intrusive while it does not directly enhance the purpose on a parent module. Remember the Unix way of doing things: one program that does a job and does it well.

5. Reading 'Claude Code' has been disturbing to me.

Please consider enriching the SlicerVMTK extension with another module that has a well defined purpose.

aabrown100-git:
@chir-set Thanks for the thoughtful review. Following your suggestion, I moved feature into a standalone CenterlineJunctionAngles module and updated the documentation accordingly. The module still uses VMTK branch extraction and bifurcation vectors internally, but it now presents the workflow as junction-angle analysis rather than as part of centerline disassembly.
I also made a few follow-up changes while testing on a large clinical centerline:
- cached branch extraction and bifurcation-vector results when the input centerline is unchanged
- added progress/status messages for the slow first-run VMTK branch extraction step
- replaced one-node-per-angle display with compact grouped model/label nodes to make large scenes faster to save/delete
- grouped annotations by pair type and branch order, with corresponding bifurcation vectors in each group

Please let me know if the standalone module organization looks right now.

chir-set:
Thank you for the rework.

I tested with `Slicer 5.12.4` on Linux and your module delivers.

I would just have one non-blocking question : why not use `Qt Designer` for the UI?

I am ready to merge it, but I would prefer an approval from @lassoan before committing to the extension.

P.S: _Out of mere curiosity, how far do you evaluate the percent contribution of Claude from the very start to now ?_

chir-set:
Would it be possible to create an icon for the module also? It's of course optional.

aabrown100-git:
@chir-set Thanks for testing and your suggestions. I’ve moved the UI into a Qt Designer file and added a module icon.

Regarding use of AI, 100% of the code was generated with either Claude or Codex. I drive the development by providing feature requests, implementation suggestions, and QA testing.

chir-set:
> into a Qt Designer file and added a module icon.

Great



> 100% of the code was generated with either Claude or Codex

Ok, this may explain
 - the use of `@staticmethod` so often, without apparent justification,
 - the use of `"nan"`, which I see for the very first time in a module,
 - the definition of the `angleDegrees()` function, while leaving aside `vtkMath::AngleBetweenVectors()`.

@lassoan, the method of creation of this module raises the question of accountability and future maintenance. I request your thoughts while being ready to merge it, or you may go ahead yourself.

It just works nicely however.

chir-set:
@aabrown100-git 

Would you consider adding a scalar array in the arc polydata? This allows a better distinction of the angles in a complex bifurcation. Coulouring the rays is not really helpful since they may overlap. Thanks.

JunctionAngles_Arcs_With_Colours

aabrown100-git:
@chir-set Thanks for the suggestion. I have made some improvements to angle visualizations, including a scalar angle array on the arc polydata, in a follow up PR (https://github.com/vmtk/SlicerExtension-VMTK/pull/196).

lassoan:
Sorry for being late here. I loosely followed the discussion and agreed where things were going, just did not have the time to write here.

Creating a separate module was the right call - it is better no to mix an editing module with a quantification module.

> the use of "nan", which I see for the very first time in a module

I would not recommend to use `nan` for error handling. A `nan` can silently propagate very far from where the error occurred and cause problems much later. It makes it very difficult to track down the root cause of the problem. Instead, if there is an invalid input then let the computation fail with an exception and then catch the exception at the appropriate level.

> the definition of the angleDegrees() function, while leaving aside vtkMath::AngleBetweenVectors().

This is a very typical AI coding error. An agent happily reimplements many existing methods. You can ask your AI agent to use existing VTK functions for basic computations and point specifically to `angleDegrees`. The problem is obviously not the implementation - that is cheap now - but consistency and maintenance (code that is not implemented does not have to be maintained).

I would add that angle measurements should use markup angles. They work better in slice views, you can interact with them, self-contained (one node for showing the angle value, interact, and store the measurement result). If some feature is missing in the angle widget then we can implement it (e.g., we may not want to show the intersection point).

PR #196, opened 2026-09-14
Improve junction angle coloring and annotation display controls

aabrown100-git:
Junction-angle arcs and labels are difficult to distinguish at complex junctions. This follow-up to #195 adds `AngleDegrees` point scalars to the arc polydata, colors arcs and labels with Plasma over a shared measured-angle range, and staggers arcs outward with labels at their midpoints to reduce overlap. Annotation rays extend as needed without changing the measured bifurcation vectors.

The module now provides visibility controls for arcs, labels, rays, vectors, parent-child and child-child pairs, and branch orders, plus ray and vector color pickers. A minimum-angle filter hides labels, arcs, and rays below the threshold without changing the measurements. Display settings are saved with the scene.

This also addresses review feedback from #195:

- invalid inputs and malformed VMTK output structures now raise explicit `ValueError`s instead of producing `NaN` measurements. Individual VMTK branch vectors with no valid direction or projected angle are skipped, and a bifurcation is skipped only if fewer than three valid branches remain;
- the custom vector-angle implementation was replaced with `vtkMath.AngleBetweenVectors`;
- native angle markups were re-evaluated and documented. They are useful for a small number of interactable angles, but one node per angle remains expensive for large trees and does not provide the display controls used here.

Markup benchmark: in Slicer 5.12.3, a synthetic 900-angle scene took about 456 ms per rendered frame, 85 s to save, and 7.3 s to delete using one native angle markup per angle. The grouped tube display took about 96 ms per frame, 0.95 s to save, and 0.5 s to delete. Native angle markups also render arcs/rays as tubes, so they do not resolve the lit-arc versus unlit-label color difference, and their arc radius is tied to the shorter ray rather than an independently staggered radius.

Validation: full `CenterlineJunctionAnglesTest().runTest()` passed in Slicer 5.12.3, including the Y-shaped tube VMTK extraction path, malformed VMTK-output validation, degenerate branch-vector skipping, scalar propagation, arc/label color matching, clustered and equal-angle color ranges, threshold filtering and restoration, display controls, and staggered trifurcation geometry. `python3 -m py_compile CenterlineJunctionAngles/CenterlineJunctionAngles.py` and `git diff --check` passed.

Colors share one range within each computation; separate computations can use different ranges. Staggering reduces crowding but cannot eliminate overlap from every viewing angle.


chir-set:
Hello,

The  angle value scalar on the arcs is great, with the Plasma colour table.

The colour table for the rays is harder to understand: two entries in the table with a constant value. They all appear in the same colour.

It would be a good enhancement to present a cleaner UI by grouping those widgets you just added in a collapsible button, closed by default.

After hitting 'Apply angle threshold', no highlighting happens anymore.

EDIT: the rays are of different colour indeed.

chir-set:
The default presentation of the result is definitely much better.

If you can pack the new control widgets in a collapsible button, I'll commit next.

JunctionAngles_Arcs_With_Colours_2

aabrown100-git:
@lassoan Addressing your comments (https://github.com/vmtk/SlicerExtension-VMTK/pull/195#issuecomment-5659338646) here.

I've implemented your suggestions about using `vtkMath.AngleBetweenVectors` and removing`NaN` values for errors.

Regarding using markup angles, I indeed had originally used markups, but found that for scenes with many annotations (on a large PA tree for instance), interaction/saving/deleting is unacceptably slow. So I replaced markup angles with grouped polydata arcs/rays and text labels. Is there a way to make angle markups more efficient? If so, happy to explore this further.

aabrown100-git:
@chir-set, thanks for these suggestions. Here are my responses

> Hello,
> 
> The angle value scalar on the arcs is great, with the Plasma colour table.
> 
> The colour table for the rays is harder to understand: two entries in the table with a constant value. They all appear in the same colour.

Good catch, the rays hold a scalar angle array so that they can be thresholded out of the scene. They intentionally all appear the same color, since rays typically overlap and coloring rays by angle would look strange. Let me know if you'd like a change regarding this. I think it might be fine to leave as is, since the actual color bar isn't shown by default. I think this confusion would only arise if the user manually shows the color bar, right?

> 
> It would be a good enhancement to present a cleaner UI by grouping those widgets you just added in a collapsible button, closed by default.

Done!

> 
> After hitting 'Apply angle threshold', no highlighting happens anymore.

You are correct, I modified this button to hide angles less than threshold, rather than highlight angles over the threshold, since angles are now colored by angle value. Let me know if you'd like to see highlighting again, perhaps through an annotation background color or box around the angle value?

> 
> EDIT: the rays are of different colour indeed.

chir-set:
> They intentionally all appear the same color, since rays typically overlap and coloring rays by angle would look strange. Let me know if you'd like a change regarding this.

No, it's good this way, a single colour for all rays since they overlap. People who want to play with the scalar can still do it on their own.

> I modified this button to hide angles less than threshold,

Ok, I missed this subtlety.

> Let me know if you'd like to see highlighting again

No, it's not that important a feature, and the threshold by angle is sufficient.

> So I replaced markup angles with grouped polydata

To me, markups are mainly for interactivity, measurements, delimiting a volume, an area ... editing something. For displaying results here, I think the polydata solution is good.

For you next PR, I would suggest keeping them as draft requests to avoid an accidental commit. When you are ready, just raise a flag.

The UI is much cleaner now, at least to me. This is just a matter of preference, you may have your preferences of course. If you look at the built-in module widgets and in extensions, these collapsible areas are used almost constantly. It's a good way to organise and present widgets in logical way for some task. Imagine the Models widget without those.

Waiting for the flag.

aabrown100-git:
@chir-set Thanks, the latest updates are pushed and I think this PR is ready for review/merge.

lassoan:
> Regarding using markup angles, I indeed had originally used markups, but found that for scenes with many annotations (on a large PA tree for instance), interaction/saving/deleting is unacceptably slow. So I replaced markup angles with grouped polydata arcs/rays and text labels. Is there a way to make angle markups more efficient? If so, happy to explore this further.

Markup display speed is fixed in the latest Slicer Preview Release. Please test and let me know if you still experience slowdowns. We can also lock the markups to remove them from picking check, which can further speed things up. If adding many markup nodes was slow: when you add more than 10-20 nodes then it may worth enabling batch update mode on the scene, make all the changes, and then disable batch mode (starting batch mode stops all incremental GUI updates and stopping it triggers full GUI update).

In general, if you run into any issues or missing features in Slicer core, please let me know. Even if we end up deciding to fix something with a temporary workaround or prototype some new features in a Python module, Slicer core developers need to be aware of what's happening, so that the needs can be addressed in upcoming updates.

aabrown100-git:
@lassoan Thanks for your comments. I just made a PR replacing the grouped polydata angle and vector annotations with native markups. In the PR, I also made requests for new features in angle markups that would improve the presentation of these junction angles.

PR #197, opened 2026-09-15
Carry a vessel end's name from Clip Vessel to the volume mesh, and fix the cap ids a boundary layer permutes

aabrown100-git:
Carries a vessel end's *name* from Clip Vessel to the volume mesh, and fixes a bug found on the
way that was permuting the cap face ids. Three separable pieces; the bug fix stands alone.

## 1. A boundary layer permuted the cap face ids (bug)

With **Boundary layer** on and **Layer on caps** off, every cap carried the id of a *different*
vessel end. The ids were the expected set — wall 1, caps 2…n+1 — so nothing looked wrong. A cap's
id becomes a `mesh-surfaces/` file name and a solver's `Add_BC` name, so an inflow waveform meant
for one vein lands on another, and nothing downstream can notice.

**Cause.** Caps are remade on the inner surface the sweep leaves behind. `BoundaryLabels` survives
that; `BoundaryPointOrder` does not — the inner boundaries are coarser than the ones the order was
written for. `GetOrExtractBoundaries` refuses the *pair* whole, leaving the capper to number caps
in its extractor's order, which is a permutation of the labels.

**Fix.** `labelsFromBoundaryPoints` reads each boundary's label off the labels its own points
carry; `nameCapsAfterTheirVesselEnd` hands the ids to the capper explicitly. A boundary whose
points say nothing, or disagree, is left to the capper rather than guessed at.

Isolated by running the stages separately — the labels, the capper alone, and remeshing were all
correct, and the layer was the only variable that changed the answer:

| checked geometrically | before | after |
|---|---|---|
| aorta, no layer / layer over caps | 0 of 3 wrong | 0 of 3 |
| aorta, layer off the caps | **3 of 3 wrong** | **0 of 3** |
| 23-cap clinical case, layer off the caps | **22 of 23 wrong** | **0 of 23** |

Reproduces on `input/aorta-surface-clipped-uncapped.vtp`, already in
[vmtk-test-data](https://github.com/vmtk/vmtk-test-data) — nothing to add there.
`test_CfdMeshGeneratorCapIdentity.py` fails without the fix on that file and passes with it.

## 2. Recording where a face's name lives

The names already exist as clip point labels, but nothing on the output pointed at them, so they
were retyped after meshing. Clip Vessel now records the pointer on its output: a `ClipPoints` node
reference, `ClipVessel.FaceIdToClipPointID`, and `ClipVessel.WallFaceID`. CFD Mesh Generator copies
all three onto the mesh. Only the pointer travels, so renaming a clip point renames the face.

Two decisions worth review:

- **Keyed by control point ID, not index.** A cap's id is `firstCapFaceId + clip point index`, so
  deleting a point would shift later names onto their neighbours' faces. By ID, a deleted point's
  face comes out *unnamed* instead.
- **Recorded where the boundaries are labelled, not where the faces are** — a boundary's label
  already *is* its cap's face id, so the mapping exists before there is a cap or a cell array. A
  workflow that leaves capping to CFD Mesh Generator has neither.

## 3. Naming the clip points, plus two smaller things

**Advanced → Vessel names**: a row per clip point, name editable in place, selecting a row selects
the point in the 3D views. Naming wants doing during clipping, not after a 15-minute mesh.

**`CapOutputSurface` now defaults to false** — the checkbox was already drawn unchecked while the
parameter default said true.

**A geometric guard.** `copyNameSource` checked only that recorded ids *exist* on the mesh, which a
permutation satisfies — hence the bug above going unnoticed. `checkCapsSitAtTheirClipPoints` now
checks the record against the geometry, matched one to one. A warning, not a refusal: a flow
extension moves a cap off its clip point and can read as a disagreement when nothing is wrong.

## Tests

32 new, across `test_CfdMeshGeneratorCapIdentity` (5), `test_CfdMeshGeneratorNameSource` (11),
`test_ClipVesselNameSource` (7) and `test_ClipVesselClipPointNames` (9). The three files written
for the record were originally id-based and would have passed on a permuted mesh; they are
geometric now, and one corrupts a record and *requires* the check to fail.

`c773303` fixes a whole-file CRLF change I made to a test `CMakeLists.txt`.

## Draft

The four commits behind the naming table are three corrections on one feature and want squashing.
Consumed by SimVascular/SlicerSimVascular#5, which depends on this.


lassoan (approved):
Thank you, it looks good. We could further improve this (terminology table, etc.) in follow-up commits.

PR #198, opened 2026-09-16
Make the bifurcation vectors the editable part of a junction angle, and fix branch order

aabrown100-git:
Follow-up to #195 and #196. This makes the bifurcation vectors the editable part of a junction angle, fixes a branch-order regression that #196 introduced, and fixes four faults found on the way.

Per @lassoan's note on #196 that markup display speed is fixed in the current Preview: the annotations are native markups throughout now, node creation runs inside `BatchProcessState`, and the angle markups are locked so they stay out of picking. Measurements below are from **Slicer 5.12.3** (macOS, r34627) and are the "before" for a Preview comparison, which I have not run yet.

## Branch order was wrong since #196

Both the depth of a branch and the parent-child graph it came from were read from the bifurcation vectors. #196 began skipping a branch whose vector VMTK could not orient, and a bifurcation left with fewer than three of them. Every branch below such a hole lost its parent, `branchOrders.get(groupId, 0)` then called it a root, and the orders below it counted up from zero again.

On a pulmonary artery tree with 111 bifurcations, where a few dozen are skipped:

| | merged `master` | this PR |
| --- | --- | --- |
| branches with the wrong order | **111 of 250** | — |
| branches reported as roots | **14** | 1 inlet (+21 one-cell stub centerlines) |
| depth histogram peak | orders 1–5 | orders 6–8 |

Depth is now taken from the ordered tracts of the extracted centerline, which describe the tree whether or not a vector could be oriented. A branch that no tract reaches is reported as **Unassigned branch order** rather than as a root.

`BranchOrder` is now the generation of the bifurcation — the order of its parent branch — so all the angles of a junction share a folder and hiding an order hides whole junctions. It was `max` of the two branch orders before, which put a junction's child-child angle one folder away from its parent-child angles. **This changes the meaning of an existing table column.**

## The vectors are what you edit

Both ends of a bifurcation vector are centerline positions, which is what makes them the right handles. An angle ray is only the measured direction drawn out from the reference-system origin, so its far end lies nowhere in particular. Each branch of each bifurcation now gets one line markup in a `Bifurcation vectors` folder; dragging either end recomputes that branch's direction and moves every angle measured from it. Angle markups are locked results. A branch is shared by the pairs of its bifurcation, so its vector exists once rather than once per pair type — 233 of 346 were drawn twice before.

A dragged endpoint can be held on the centerline, and is offered **its own branch and no other**: a search of the whole tree pulls a handle onto whichever branch lies nearest in space, which in a dense tree is usually not the one under the cursor. Which part of that branch is taken is decided **along the line of sight**, because a drag in a 3D view slides a handle across a plane facing the camera, so the handle is under the cursor; choosing the part nearest in space picks whatever is closest in depth and the handle appears to ignore the cursor.

Over 2070 simulated 3 mm drags on that tree, the snap was refused:

| | refused |
| --- | --- |
| nearest in space, reach of 3 vector lengths | 47.1% |
| along the line of sight, reach of 3 | 25.9% |
| along the line of sight, reach of 6 (used) | **1.9%** |

The centerline of the branch is stored on the vector, so a reopened scene behaves the same way.

## Faults fixed

- `onResetSelectedAngle` bound `_` as a loop variable and then called it as the translator, so the button raised `TypeError` on the common path where nothing is selected. `computeBranchOrders` did the same to its array validation, turning a `ValueError` into `UnboundLocalError`.
- The branch identity written to the node description never survived: Slicer rewrites a description from the measurements.
- Angle markups were observed and never released — 748 observations for 372 nodes, holding deleted nodes alive.
- The moved-point handler rebuilt a lookup table and rewrote the visibility of every displayable node in the scene on **every mouse move**: 6.8 ms an event where 0.034 ms is enough.
- A vector keeps its two ends. One in my scene had lost both control points and could not be repaired by hand, so the ends can no longer be deleted and a stripped vector is rebuilt from the ends recorded on it.

## Limitations

- **Rendering, on 5.12.3.** With the whole tree shown (372 angles + 346 vectors) a frame takes ~405 ms; angles only, ~156 ms. A drag event costs almost exactly one frame, so the module's own work is not the bottleneck. The threshold and branch-order filters keep the visible count down in practice. Needs re-measuring on the Preview.
- **Node count.** One markup per angle plus one per branch vector; 718 markup nodes on this tree.
- **The label reads `Angle: 47.7°`.** See below.
- **Colours are comparable only within one computation**, since the range is the measured minimum and maximum of that run.
- **21 one-cell stub centerlines** on this tree become their own order-0 roots. They take part in no measured angle, but they are Extract Centerline artifacts rather than real trees.
- **The snap has no screen-space notion of reach.** It is measured in world mm at the depth of the branch, so it does not scale with zoom. Pixels would be the right unit.
- The `parent-parent` pair type is reachable but untested against real data, and one bifurcation on this tree has no parent branch at all; its angles fall back to the shallowest branch order present.

## What we would like changed in Slicer core

All observed in 5.12.3 and reproducible; happy to file these separately.

**1. An option to leave the node name out of the properties label.** The rendered label is always `<node name>: <enabled measurements>`, so a generated annotation reads `Angle: 47.7°`. Renaming to `XY` gives `XY: 73.9°`; an empty name gives a bare `: 35.7°`. For generated results the name is redundant — identity is in the folder tree and the node attributes — and it costs roughly half the label width when hundreds are on screen. Something like `SetPropertiesLabelIncludesNodeName(bool)` on the display node would be enough. The workaround (disable the measurement, put the value in the node name) renders correctly but drops the angle from the Markups module's measurement list and leaves the name a copy that must be kept in sync, so it is not used here.

**2. `SetPropertiesLabelVisibility(false)` appears to be ignored.** Setting it false on an angle markup leaves the label drawn; `GetPropertiesLabelVisibility()` reads back false in the same state. This looks independent of 1.

**3. `GetPropertiesLabelText()` does not match what is rendered.** For a node named `Angle` measuring 137.5° it returns `" 137.5°"` — measurement only, no name, no colon — while the view draws `Angle: 137.5°`. The name is prepended in the representation rather than in that method.

**4. A pickable properties label.** The pickable components are the control points, the line, the centre point, the plane and the interaction handles; there is no component for the label, so hovering the text cannot highlight or select its markup.

**5. Selectable but not editable.** Locking a markup removes it from picking entirely, so a locked result cannot be hovered or clicked. Leaving the node unlocked and locking its control points individually does give hover and click while keeping it non-editable, but it is an indirect way to say it, and on 5.12.3 it takes mouse-move handling from 143 ms to 371 ms with the whole tree shown.

## Validation

Full `CenterlineJunctionAnglesTest().runTest()` passes in Slicer 5.12.3, run in a separate `--testing --no-main-window` process. New coverage: the generation of a junction, a vector driving the angles of its branch, resetting a vector, snapping to the branch under the cursor rather than to a decoy 0.1 mm away in space, the reach of the snap, the line of sight against a limb nearer in space, a vector's ends surviving deletion, the lifetime of the observations, and the validation of the extracted arrays. `python3 -m py_compile` and `git diff --check` pass.

Draft while the Preview comparison is outstanding.


aabrown100-git:
@lassoan @chir-set could you review when you get a chance?

lassoan (approved):
Thank you, it looks good to me. We'll follow up on Slicer core changes.

lassoan:
Slicer core changes:

- 1-3 should be no problem, please send a PR if you can. If not then submit an issue and I'll get to it soon.
- 4. A pickable properties label. It is not easy, but there have been some [draft implementations](https://github.com/Slicer/Slicer/pull/6809) and @sunderlandkyl worked on it recently. We can try  to bring it up in the priority queue.
- 5. Selectable but not editable. This is already available by locking every control point (instead of locking the whole markup).

aabrown100-git:
@lassoan Thanks, I made a PR (https://github.com/Slicer/Slicer/pull/9410) addressing these issues.

